PR #1, opened 2016-02-20
Formatting + java mvn lang level prop

rzwiefel:
Formatted poms so they don't look like trash - sorry it bugged me.
Also found the maven compiler property.. which actually makes your project set the right module version when you import it into intelliJ.

@andrewkfiedler


lessarderic:
:shipit:

PR #10, opened 2016-07-19
Fix polygon height so polygons are drawn at the same height as the corresponding billboards

jrnorth:
#### What does this PR do?

Fixes polygons' draw height in Cesium so polygons are drawn at the same height as the corresponding billboards.
#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged; if a component team is listed, at least one of its members needs to approve)?

@djblue 
@andrewkfiedler 
#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).

@pklinef
@jlcsmith 

Ingest some metacards/products with geospatial data, search for them on the 3d map, and verify that all polygons are drawn at the same height as their corresponding billboards.
#### Screenshots (if appropriate)

screen shot 2016-07-19 at 3 48 33 pm
#### Checklist:
- [] Documentation Updated
- [] Update / Add Unit Tests
- [] Update / Add Integration Tests

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/webapp/js/view/cesium.metacard.js
+++ b/catalog/ui/catalog-ui-search/src/main/webapp/js/view/cesium.metacard.js
@@ -429,7 +429,10 @@ define([
 
             getPolygonOutline: function(positions) {
                 return new Cesium.GeometryInstance({
-                    geometry: Cesium.PolygonOutlineGeometry.fromPositions({positions: positions}),
+                    geometry: Cesium.PolygonOutlineGeometry.fromPositions({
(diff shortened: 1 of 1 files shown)


jrnorth:
Note that in the attached screenshot the polygon is floating above the ground. This can be fixed by increasing the detail level we use for our terrain height sampling. Currently we're using level 5, but I think it goes as high as 15. I was just wondering why we're using 5 right now.

pklinef:
I am not aware of a reason why we picked 5.  We can try increasing it to 15 and see if anything breaks.

PR #22, opened 2016-10-11
Catalog UI bundle shutdown, startup cleanly. Resolved ConcurrentModificationException

lcrosenbu:
#### What does this PR do?

When shutting down the Catalog UI bundle, the operation would fail with a ConcurrentModificationException due to attempting to remove an item from a collection while simultaneously iterating over the same collection. This PR safely removes the item from the collection while iterating. 
#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged; if a component team is listed, at least one of its members needs to approve)?

@bdeining @rzwiefel @djblue 
#### Choose 2 committers to review/merge the PR (please choose ONLY two committers from below, delete the rest).

@jlcsmith 
@stustison

Build and run DDF. Stop the Catalog UI application using the Admin Console. No ConcurrentModificationException errors should occur, and the application should stop successfully. Re-start the Catalog UI application. 
#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
+++ b/catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -392,10 +393,11 @@ public class ConfigurationApplication implements SparkApplication {
     }
 
     private void stopImageryEndpoints(List<String> imageryEndpointsToStop) {
-        for (String endpoint : imageryEndpointsToStop) {
+        for (Iterator<String> iterator = imageryEndpointsToStop.iterator(); iterator.hasNext(); ) {
+            String endpoint = iterator.next();
             try {
                 httpProxy.stop(endpoint);
-                imageryEndpoints.remove(endpoint);
+                iterator.remove();
             } catch (Exception e) {
                 LOGGER.error("Unable to stop proxy endpoint: {}", endpoint, e);
             }


stustison (approved):

bdeining (approved):

peterhuffer (approved):

rzwiefel:
~~This is @djblue's code, so he should be tagged on the PR.~~ Oh nvm it was stustison looks like that  did that line. Regardless though, It would be nice to tag the original file creator on the PR since it's still kind of "his code".

edit 2: actually looks like this was just copied over from somewhere else 😐

catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
```diff
@@ -392,10 +393,11 @@ private void findDifferences(List<Map<String, Object>> innerList,
     }
 
     private void stopImageryEndpoints(List<String> imageryEndpointsToStop) {
-        for (String endpoint : imageryEndpointsToStop) {
+        for (Iterator<String> iterator = imageryEndpointsToStop.iterator(); iterator.hasNext(); ) {
+            String endpoint = iterator.next();
             try {
                 httpProxy.stop(endpoint);
-                imageryEndpoints.remove(endpoint);
+                iterator.remove();
```
rzwiefel, line 400: The `httpProxy.stop(endpoint)` shows up three separate times in this class, so it might be worth refactoring out to a method.

Then this entire method could be refactored to:

``` java
imageryEndpointsToStop.forEach(this::tryStopHttpProxy);
imageryEndpoints.removeAll(imageryEndpointsToStop);
```

rzwiefel (approved):
Approved, optional comment to resolve if your conscious dictates you to do so.

jlcsmith (approved):

bdeining:
Built and tested and was able to turn off the catalog-app without any errors  👍

PR #28, opened 2016-11-02
Fixed injected attributes not being returned in metacardtype endpoint

rzwiefel:


catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -181,6 +192,29 @@ public Metacard getMetacard(String id)
             }
             resultTypes.put(metacardType.getName(), attributes);
         }
+        for (InjectableAttribute attribute : injectableAttributes) {
```
jlcsmith, line 195: It seems like this block and the one above are very similar.  Can we move this to a common method?

rzwiefel: probably not because we're iterating over different objects  (two different interfaces).

rzwiefel: one iterates over the metacard type -> then each attribute in the type
the other iterates over each Injectable attribute -> then over each type that injectable attribute belongs in

catalog/ui/catalog-ui-search/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -124,6 +124,8 @@
     <reference-list id="metacardMultiActionProviders" interface="ddf.action.MultiActionProvider"
                     filter="(id=catalog.data.metacard.*)"/>
 
+    <reference-list id="injectableAttributes" interface="ddf.catalog.data.InjectableAttribute" />
```
bdeining: Shouldn't this be optional?  This might time out if there are no registered InjectableAttributes

rzwiefel: yes it should be. fixed.

PR #34, opened 2016-11-09
Fixed reverting update issue and added logging

rzwiefel:


catalog/ui/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -479,16 +488,46 @@ private void revertContentandMetacard(Metacard latestContent, Metacard versionMe
...
+        if (retries <= 0) {
+            throw new IngestException("Could not update metacard!");
+        }
+        LOGGER.trace("Trying to update metacard.");
+        try {
+            func.call();
+            LOGGER.trace("Successfully updated metacard.");
+        } catch (Exception e) {
+            LOGGER.trace("Failed to update metacard");
+            trySleep(350);
```
bdeining, line 523: 500ms might be better here to account for a higher latency environment

bdeining (approved):

bdeining (approved):
Hero Built and Tested

rzwiefel:
If we had more time I'd switch this out for the failsafe library which would be even better.

PR #42, opened 2016-12-08
Updated revert path to allow for web context path security

rzwiefel:


jlcsmith:
Verified that revert and restore can now be controlled via a web context policy on "/search/catalog/internal/history/revert".  However, the revert/restore action returns a 403 to the Catalog UI and results in a generic "Server Error: Unknown Error".  There should be a more appropriate error message like "Action is not authorized" or similar.

rzwiefel:
@jlcsmith Do you remember what exception was actually thrown in the logs?

jlcsmith:
Sorry @rzwiefel I don't have it handy.  It was a 403 unauthorized though.

rzwiefel:
Fixed the forbidden message in UI and did full hero build

PR #47, opened 2016-12-22
Update QueryPolling to perform retries when sources are down

andrewkfiedler:
- Whenever the source information is synced, failures will be checked to see if they correspond to a now available source.  If so, the failures are batched into one request (min and max time range).
 - If a query is updated, the latest query information is always used on a retry.
 - If a query is updated and the source corresponding to a retry is no longer part of the query, that retry is removed from the queue


pklinef:
Manually merged

PR #62, opened 2020-01-22
Resolve transitive dependency on sts

rzwiefel:


Bdthomson:
✅ Hero Successful

image

 
Verified:
- Feature was able to install on DDF-2.22.0-SNAPSHOT.

PR #67, opened 2020-01-27
DDF-5627 G-7443 Add an endpoint to validate different aspects of a query

rzwiefel:


rzwiefel:
Verified can hit the new endpoint when deploying the bundles to DDF-2.23.0-SNAPSHOT.
Merging since this already passed review in ddf repo.

PR #74, opened 2020-01-28
DDF-5360 Changed user ID based OAuth token storage to session ID to better handle the guest user.

blen-desta:
Previously, the user's email address or username (if an email address is not present) was used to store users' OAuth tokens. This would cause problems for guest users. To solve that, we're switching to store tokens based on session IDs.

Related to: https://github.com/codice/ddf/pull/5682

### DIFF
--- a/ui-backend/catalog-ui-oauth/src/main/java/org/codice/ddf/catalog/ui/oauth/app/OAuthApplication.java
+++ b/ui-backend/catalog-ui-oauth/src/main/java/org/codice/ddf/catalog/ui/oauth/app/OAuthApplication.java
@@ -23,7 +23,6 @@ import static org.codice.ddf.security.token.storage.api.TokenStorage.DISCOVERY_U
 import static org.codice.ddf.security.token.storage.api.TokenStorage.SECRET;
 import static org.codice.ddf.security.token.storage.api.TokenStorage.SOURCE_ID;
 import static org.codice.ddf.security.token.storage.api.TokenStorage.STATE;
-import static org.codice.ddf.security.token.storage.api.TokenStorage.USER_ID;
 import static spark.Spark.get;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -46,6 +45,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
 import java.util.Map;
+import javax.servlet.http.HttpSession;
 import org.codice.ddf.configuration.SystemBaseUrl;
 import org.codice.ddf.security.oidc.resolver.OidcCredentialsResolver;
 import org.codice.ddf.security.oidc.validator.OidcTokenValidator;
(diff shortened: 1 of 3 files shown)


ui-backend/catalog-ui-oauth/src/main/java/org/codice/ddf/catalog/ui/oauth/app/OAuthApplication.java
```diff
@@ -201,8 +202,11 @@ String processCodeFlow(Response res, String code, Map<String, Object> state) {
 
     JWT idToken = oidcTokens.getIDToken();
     AccessToken accessToken = oidcTokens.getAccessToken();
-    RefreshToken refreshToken = oidcTokens.getRefreshToken();
+    RefreshToken refreshToken = null;
 
+    if (!userId.startsWith(GUEST)) {
```
SmithJosh: Edge case, but maybe `userId.equals(GUEST)` would be better: https://en.wikipedia.org/wiki/Guest_(surname)

blen-desta: The userId is usually `Guest@<IP>` that's why I was using `startsWith`

bakejeyner (approved):

ui-backend/catalog-ui-oauth/src/main/java/org/codice/ddf/catalog/ui/oauth/app/OAuthApplication.java
parret: If it's always in that format, maybe including the `@` in the `startWith` check could catch that edge case

parret (approved):

ui-backend/catalog-ui-oauth/src/main/java/org/codice/ddf/catalog/ui/oauth/app/OAuthApplication.java
SmithJosh: ❗️ I tested this and was able to use `SubjectUtils.isGuest(...)`. That would be the best option

ui-backend/catalog-ui-oauth/src/main/java/org/codice/ddf/catalog/ui/oauth/app/OAuthApplication.java
```diff
@@ -201,8 +202,11 @@ String processCodeFlow(Response res, String code, Map<String, Object> state) {
...
+      refreshToken = oidcTokens.getRefreshToken();
```
SmithJosh: ❗️ refreshToken will be null if user is guest. That causes an NPE on line 220

SmithJosh (changes requested):

ui-backend/catalog-ui-oauth/src/main/java/org/codice/ddf/catalog/ui/oauth/app/OAuthApplication.java
```diff
@@ -201,7 +203,11 @@ String processCodeFlow(Response res, String code, Map<String, Object> state) {
 
     JWT idToken = oidcTokens.getIDToken();
     AccessToken accessToken = oidcTokens.getAccessToken();
-    RefreshToken refreshToken = oidcTokens.getRefreshToken();
+    String refreshToken = null;
+
+    if (!isGuest()) {
+      refreshToken = oidcTokens.getRefreshToken().getValue();
```
bdeining: it looks like there is a potential NPE here: https://www.javadoc.io/doc/com.nimbusds/oauth2-oidc-sdk/latest/com/nimbusds/oauth2/sdk/token/Tokens.html#getRefreshToken()

stustison (approved):

parret (approved):

bakejeyner (approved):

SmithJosh (approved):

SmithJosh:
Hero successful 🎉 (tested using the checklist on the linked DDF PR)

PR #78, opened 2020-01-29
Lock jquery and backbone versions

rzwiefel:


bdeining:
i'm assuming we can close https://github.com/codice/ddf-ui/pull/49?

willwill96:
There should probably be a yarn.lock change associate with this

willwill96:
hero successful, pulled down and verified ui loaded both from dev server and port 8993 after installing features

PR #82, opened 2020-02-03
Update version for catalog-ui-search in package.json to match released version

Bdthomson:


ui-frontend/packages/catalog-ui-search/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "catalog-ui-search",
   "author": "Codice",
   "description": "A frontend UI for DDF Catalog",
-  "version": "0.0.1",
+  "version": "3.0.0",
```
bdeining, line 5: should this be the next released version?

Bdthomson: I didn't know so I just checked a popular project (Material UI) and they have it set how I do - because there is no such thing as a snapshot version in js, the current version matches what's in master.

image
 https://material-ui.com/ and https://github.com/mui-org/material-ui/blob/master/package.json#L2

bdeining (approved):

PR #90, opened 2020-02-06
Use SelectionInterface in metacard overwrite

vinamartin:
https://github.com/codice/ddf-ui/issues/89

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/metacard-overwrite/metacard-overwrite.view.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/metacard-overwrite/metacard-overwrite.view.js
@@ -13,7 +13,7 @@
  *
  **/
 
-const store = require('../../js/store.js')
+const SelectionInterface = require('../selection-interface/selection-interface.model.js')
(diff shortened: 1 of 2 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/metacard-overwrite/metacard-overwrite.view.js
```diff
@@ -211,7 +211,7 @@ class MetacardOverwrite extends React.Component {
   constructor(props) {
     super(props)
     this.state = defaultState
-    this.model = store.getSelectedResults().first()
+    this.model = SelectionInterface.getSelectedResults().first()
```
willwill96, line 214: For this to work, I think you need to pass an instantiated selection interface model from the parent component. I don't think you can globally access `SelectionInterface` like this

vinamartin: You're right, I was having trouble testing this one but as soon as I got it running I realized this wouldn't work.

vinamartin:
Need to pass in via props, closing this PR for now

PR #101, opened 2020-02-19
Upgraded geotools versions in line with ddf

rzwiefel:
### This will need the DDF 2.23.0 Release version before we can merge.

This is the immediate fix. We are looking into a longer term solution that does not couple us so tightly to ddf's change.


ui-backend/catalog-ui-search/pom.xml
```diff
@@ -158,19 +158,24 @@
         <dependency>
             <groupId>org.geotools</groupId>
             <artifactId>gt-opengis</artifactId>
-            <version>${opengis.version}</version>
```
bdeining, line 161: why does this have to change to a hardcoded version?

rzwiefel: because there's only a single usage of it, the property probably should have been removed but looks like i left it in there.

bdeining: there's more than 1 usage of it though; it seems like we should use the property and revert these changes

PR #108, opened 2020-03-04
Fix intrigue security

rzwiefel:
## Will need the DDF-2.23.0 release before merging

⚠️ The upgrade geotools commit will disappear when that PR is merged.

Changes Intrigue to use the new ddf security interfaces


ui-backend/catalog-ui-search/pom.xml
```diff
@@ -158,19 +158,24 @@
         <dependency>
             <groupId>org.geotools</groupId>
             <artifactId>gt-opengis</artifactId>
-            <version>${opengis.version}</version>
+            <version>20.1</version>
```
bdeining, line 161: it's not clear why we wouldn't want to use the root level `opengis.version` here

PR #109, opened 2020-03-05
Rename src to sources for RPC

stevenmalmgren:
The front end stores the sources on a query metacard in a property called src. When submitting with an RPC request the src property will be ignored as it doesn't match the "sources" definition on the metacard.


jordanwilking (approved):

bdeining (approved):

vinamartin:
✅ Manually verified

vinamartin (approved):

PR #120, opened 2020-03-18
DDF-5900 update for security api and impl split

stustison:


bdeining:
are we planning on moving to snapshot or is that temporary?

pklinef (approved):

jrnorth (approved):

pom.xml
```diff
@@ -51,7 +51,7 @@
     <maven.deploy.plugin.version>2.8.2</maven.deploy.plugin.version>
 
     <!--Backend properties-->
-    <ddf.version>2.23.0</ddf.version>
+    <ddf.version>2.24.0-SNAPSHOT</ddf.version>
```
rzwiefel, line 54:
```suggestion
    <ddf.version>2.24.0</ddf.version>
```
This suggestion will make sure CI wont pass until 2.24.0 released, which is probably preferable now that we've tested that it works with the snapshot version.

jlcsmith:
close to https://github.com/codice/ddf-ui/pull/209 ?

rzwiefel:
Closed per #209 as mentioned above

PR #151, opened 2020-03-31
DDF-UI-145 default result form

### LINKED ISSUE #145: User should be able to select a default Result Form
Currently, the user can only select a default Search Form.  This functionality should be extended to result forms.  If the user sets a default result form, that default should appear as the default in the Search Form builder, as well as the advanced and basic search forms.

cassandrabailey293:
Forward port of https://github.com/codice/ddf/pull/5957

___________________________________________________

#### What does this PR do?
Adds the ability to set a default result form. 

Setting a result form as default means that it will show up automatically in these cases:
1. When you create a basic or advanced search in your workspace
2. When you create a new search form in the search form editor

If you set a result form to default, but you already have a default search form that does not have that result form, you will NOT see the default result form when you search. You will see the result form associated with your default search form. 

#### Who is reviewing it? 

@andrewzimmer @hayleynorton @zta6 

@codice/ui 

1. build/profile install
2. create two result forms if you don't already have them
3. Verify that the dropdown has the "Set Default" option
4. Set one as the default
5. Verify that the result form you set as default appears in the basic and advanced search (make sure you don't have a default search form set)
6. Verify that the result form you set as default appears when you're creating a new search form
7. Create two search forms, one with one of your result forms, and another with no result form specified ("All Fields")
8. Verify that when you set these search forms to default, the default result form doesn't show up in the Basic and Advanced search. Rather, the result form you set ON the search from should show up. 

Fixes: https://github.com/codice/ddf-ui/issues/145

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
@@ -93,10 +93,11 @@ module.exports = plugin(
       })
       resultTemplates = _.uniq(resultTemplates, 'id')
       let lastIndex = resultTemplates.length - 1
+      let defaultResultForm = resultTemplates.find((form) => form.id === this.model.get('defaultResultFormId'))
       let detailLevelProperty = new Property({
         label: 'Result Form',
         enum: resultTemplates,
-        value: [
+        value: [ defaultResultForm && defaultResultForm.value ||
           this.model.get('detail-level') ||
             (resultTemplates &&
               resultTemplates[lastIndex] &&
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
@@ -93,11 +93,13 @@ module.exports = plugin(
       })
       resultTemplates = _.uniq(resultTemplates, 'id')
       let lastIndex = resultTemplates.length - 1
-      let defaultResultForm = resultTemplates.find((form) => form.id === this.model.get('defaultResultFormId'))
+      let defaultResultForm = resultTemplates.find(
+        form => form.id === this.model.get('defaultResultFormId')
+      )
       let detailLevelProperty = new Property({
         label: 'Result Form',
         enum: resultTemplates,
-        value: [ defaultResultForm && defaultResultForm.value ||
+        value: [ (defaultResultForm && defaultResultForm.value) ||
           this.model.get('detail-level') ||
             (resultTemplates &&
               resultTemplates[lastIndex] &&
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
@@ -99,8 +99,9 @@ module.exports = plugin(
       let detailLevelProperty = new Property({
         label: 'Result Form',
         enum: resultTemplates,
-        value: [ (defaultResultForm && defaultResultForm.value) ||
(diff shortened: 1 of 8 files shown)


zta6 (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
```diff
@@ -93,11 +97,15 @@ module.exports = plugin(
...
       resultTemplates = _.uniq(resultTemplates, 'id')
       let lastIndex = resultTemplates.length - 1
+      let defaultResultForm = resultTemplates.find(
+        form => form.id === user.getQuerySettings().get('defaultResultFormId')
+      )
       let detailLevelProperty = new Property({
         label: 'Result Form',
         enum: resultTemplates,
         value: [
-          this.model.get('detail-level') ||
```
andrewzimmer, line 100: I think we'll want to use the model value first here

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
```diff
@@ -93,11 +97,15 @@ module.exports = plugin(
...
+          (defaultResultForm && defaultResultForm.value) ||
+            this.model.get('detail-level') ||
```
willwill96: How do `defaultResultFormId` and `detail-level` now relate to each other?

cassandrabailey293: the `detail-level` isn't the id, it's another unique identifier. It can have the value `allFields` which the `defaultResultFormId` cannot. I could try replacing `this.model.get('detail-level')` with `allFields` - i think this would achieve the desired logic

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
cassandrabailey293: what do you mean the model value? Here's the model represents the entire search form. e.g. this search form is called `ihavearesultform` : 
image

cassandrabailey293: are you saying use the value for `defaultResultFormId` first? I was using the `detail-level` because I had an issue where if the default result form was set, but the default search form was also set, and the default search form had no result form, it was showing the default result form rather than "All Fields" when I went to search. I just tried it again and I seem to still have that issue, so I'll need to look into this

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/query-settings.view.js
willwill96: I'm still confused as to what the difference is between the two attributes are. I understand that they reference result forms in different ways (one by id and the other by name), but they both represent a default result form and the distinction is unclear to me

cassandrabailey293: I think the `detail-level` could have a result form without it being the "default" result form. Like, if aa user selects a result form while on a search form, that result form would be the `detail-level`. but the default result form is something that's set in the result form selection, and isn't necessarily the search form's result form. 

In other words.. search form A could have result form X, and you've made search from A your default search form. Then, you decide to make result form Y your default result form. When you go to search, it will default to search form A, which will show X rather than Y. In this case, X is the `detail-level` and Y is the `defaultResultFormId`

hayleynorton (approved):

bdeining (approved):

bdeining:
hero'd here https://github.com/codice/ddf/pull/5957

Issue #185, opened 2020-04-29
Configurable set of metacard tags for enforcing access controls


jzstern:
We need to be able to configure additional tags for metacards that aren't complying with catalogue's access control

PR #187, opened 2020-04-29
Add metacard tag configuration for enforcing access controls

jzstern:
This PR adds configuration for additional metacard tags to comply with catalogue's access control

Addresses #185

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/resources/OSGI-INF/metatype/org.codice.ddf.catalog.ui.security.accesscontrol.xml
+++ b/ui-backend/catalog-ui-search/src/main/resources/OSGI-INF/metatype/org.codice.ddf.catalog.ui.security.accesscontrol.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+/**
+ * Copyright (c) Codice Foundation
+ *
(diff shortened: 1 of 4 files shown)


criscrasbell (approved):

ui-backend/catalog-ui-search/src/main/resources/OSGI-INF/metatype/org.codice.ddf.catalog.ui.security.accesscontrol.xml
```diff
@@ -0,0 +1,30 @@
...
+ *
+ **/
+
+ -->
+<metatype:MetaData xmlns:metatype="http://www.osgi.org/xmlns/metatype/v1.0.0">
+
+    <OCD name="Intrigue Metacard Tag Config" id="org.codice.ddf.catalog.ui.security.accesscontrol">
+        <AD id="intrigueTags"
+            name="Intrigue Tags"
+            type="String"
```
jlcsmith, line 21: Need to add a description here.  What is this config doing?  What does it control?

Lambeaux (commented):
Would strongly prefer avoiding the exposure of more configuration and go with the more generic metacard tag, such as `policy-shareable`, instead of specific ones like `workspace`, etc; however, if we have no other alternative I don't mind looking at this in more depth and potentially approving.

bdeining (approved):
lgtm after final comments

PR #192, opened 2020-05-01
DDF-5649 G-6176 G-7742 Show warnings to the user

kentmorrissey:
**What does this PR do?**

This PR:

- [adds a method called `getWarnings` to the `CqlQueryResponse`](https://github.com/codice/ddf-ui/pull/192/files#diff-dc2bbd07808579ad832d96271a7611f5R36-R37) so that the application [can add to the `CqlQueryResponse` any user-friendly warning which it has generated from the query before it serializes the `CqlQueryResponse` to the UI](https://github.com/codice/ddf-ui/pull/192/files#diff-770098fd5460b75665d959794e61fd7bR127)

- [implements `getWarnings` in the `CqlQueryResponseImpl`](https://github.com/codice/ddf-ui/pull/192/files#diff-db7a5b9d957f1ebb2d1bc9b662743b0fR35-R215)

- [creates a `SourceWarningsFilterManager`](https://github.com/codice/ddf-ui/pull/192/files#diff-f0034ba72b72d296732386e06ed2d5a5R1-R30) which [aggregates the `SourceWarningsFilters` which the application makes available](https://github.com/codice/ddf-ui/pull/192/files#diff-4778db6e796b30e349fc379c38b1c54aR324-R333) so that the `SourceWarningsFilterManager` [can filter the warnings of `ProcessingDetails` which it receives such that the `ProcessingDetails` only retain those warnings which the best compatible `SourceWarningsFilter` deems user-friendly](https://github.com/codice/ddf-ui/pull/192/files#diff-f0034ba72b72d296732386e06ed2d5a5R31-R65)

- [adds a `SourceWarningsFilterManager` to the `QueryApplication`](https://github.com/codice/ddf-ui/pull/192/files#diff-770098fd5460b75665d959794e61fd7bR87-R113) so that [when the `QueryApplication` generates a `CqlQueryResponse` which contains `ProcessingDetails`, the `QueryApplication` can filter as described above the `warnings` of the `ProcessingDetails` and add each user-friendly `warning` to the `CqlQueryResponse` before the `QueryApplication` serializes the `CqlQueryResponse` to the UI](https://github.com/codice/ddf-ui/pull/192/files#diff-770098fd5460b75665d959794e61fd7bR245-R267)

- [alters `catalog-ui-search`'s `blueprint.xml` such that `catalog-ui-search` listens for any `SourceWarningsFilter` which the application registers, adds each such `SourceWarningsFilter` to a list of `SourceWarningsFilter`s](https://github.com/codice/ddf-ui/pull/192/files#diff-4778db6e796b30e349fc379c38b1c54aR324-R328), [passes this list to the constructor of a `SourceWarningsFilterManager`](https://github.com/codice/ddf-ui/pull/192/files#diff-4778db6e796b30e349fc379c38b1c54aR330-R332), and [passes this `SourceWarningsFilterManager` to the constructor of the `QueryApplication`](https://github.com/codice/ddf-ui/pull/192/files#diff-4778db6e796b30e349fc379c38b1c54aR334-R345)

- [adds tests which verify the correct operation of the `SourceWarningsFilterManager`](https://github.com/codice/ddf-ui/pull/192/files#diff-5ba0e520871c0d51b1b539bc65661536R1-R191)

- [changes `QueryResponse.js` such that when it receives a `CqlQueryResponse` with non-empty `warnings`, it, for each `warning`, displays to the user an `announcement` which shows the ID of the source of the `warning` as well as the text of the `warning`
](https://github.com/codice/ddf-ui/pull/192/files#diff-c04f18cbeb1160f9a442fe48158973c3R25-R221)
- changes [each instance of `Announcment`, `accouncment`, or `Announcments`](https://github.com/codice/ddf-ui/pull/192/files#diff-15eef8e7a3f7b0b9e18a75a44eeb9e2cL1-R1) in [`catalog-ui-search`](https://github.com/codice/ddf-ui/pull/192/files#diff-19edb2527c5feec566f43c0e42cceb9eL44-R93) to [`Announcement`, `announcement`, or `Announcements`](https://github.com/codice/ddf-ui/pull/192/files#diff-bf461ec66c946c62da1f1a0180d38219L24-R33), respectively.

**Who is reviewing it?**

@Lambeaux 
@mojogitoverhere 
@brianfelix 
@brhumphe 

**Select relevant component teams:**

@codice/ui

**Ask 2 committers to review/merge the PR and tag them here.**

**How should this be tested?**

(if something goes awry with any of the steps which I list below, bug me on Slack)

- in your Terminal/Command Line/what have you, navigate to where you have your local `ddf-ui`

- run `git checkout master`

- run `git pull origin master`

- run `git fetch origin pull/192/head:ddf-5649`

- run `git checkout origin/pr/192`

- run `git checkout -b whateverYouWantToCallYourLocalBranchOfThisPR`

- run `git rebase master`

- verify that, in response, git outputs something like
`Current branch whateverYouWantToCallYourLocalBranchOfThisPR is up to date.`

- verify that, on your newly created local branch, `ddf-ui/pom.xml` contains `<ddf.version>2.24.0<.ddf.version>`

- run `mvn clean install`

- navigate to where you have your local `ddf` and, if you don't already have a local branch of https://github.com/codice/ddf/releases/tag/ddf-2.24.0, run 
`git fetch origin :remotes/origin/master`, and then `git checkout ddf-2.24.0`, and then `git checkout -b whateverYouNameYourLocalBranch`. Otherwise, run 
`git checkout whateverYouHaveNamedYourLocalBranch`

- verify that, on your local branch of `ddf-2.24.0`, `ddf/pom.xml` contains`<version>2.24.0</version>`

- message me on Slack so that I can send you `implementationOfSourceWarningsFilter.patch`

- `git apply implementationOfSourceWarningsFilter.patch` to your local branch

- build your local branch

- run `cd distribution/ddf/target && unzip ddf-2.24.0.zip && cd ddf-2.24.0/bin && ./ddf`

- if you haven't already started your Docker Desktop, do so

- in a separate tab or window of your Terminal/Command Line/what have you, navigate to where you have your local `ddf`

- run `cd distribution/docker/solrcloud`

- run `docker-compose up`

- in the tab or window of your Terminal/Command Line/what have you in which you've started the Karaf Console, run `profile:install standard`

- after `profile:install standard` has completed, navigate to https://localhost:8993/admin

- if prompted for credentials, enter the admin credentials

- create a source for OpenSearch and activate it

- go back to the window or tab of your Terminal/Command Line/what have you where you have your running instance of `ddf`

- run `feature:repo-add mvn:org.codice.ddf.search/intrigue-ui-app/3.6.0-SNAPSHOT/xml/features`

- run `feature:install catalog-ui-app`

- run `feature:repo-add mvn:org.codice.ddf.search/ui-frontend/3.6.0-SNAPSHOT/xml/features`

- run `feature:install ui-frontend`

- on a different browser than the one with which you created a source for OpenSearch, navigate to https://localhost:8993/search/catalog/

- run a wildcard search for the source which you created for OpenSearch

- an `announcement` should pop up which has the name of the source which you created for OpenSearch as its title, and "looks like the test passed!" as its message.

**Any background context you want to provide?**

This PR enables that DDF's UI display to the user any warning which
a `SourceWarningsFilter` has approved. With this enhancement, an `Object`
can communicate to the user specific, comprehensible information about a
malfunction, as well information about how the user may resolve the
malfunction.

**What are the relevant tickets?**

**Screenshots**

**Checklist:**

- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

**Notes on Review Process**

**Review Comment Legend:**

### DIFF
--- a/ui-backend/catalog-ui-search-api/src/main/java/org/codice/ddf/catalog/ui/query/utility/CqlQueryResponse.java
+++ b/ui-backend/catalog-ui-search-api/src/main/java/org/codice/ddf/catalog/ui/query/utility/CqlQueryResponse.java
@@ -16,6 +16,7 @@ package org.codice.ddf.catalog.ui.query.utility;
 import ddf.catalog.operation.QueryResponse;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * <b> This code is experimental. While this interface is functional and tested, it may change or be
@@ -32,4 +33,6 @@ public interface CqlQueryResponse {
   String getId();
 
   Status getStatus();
+
+  Set<String> getWarnings();
 }
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
@@ -22,15 +22,18 @@ import static spark.Spark.post;
 import com.google.common.collect.ImmutableMap;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import ddf.catalog.operation.ProcessingDetails;
 import ddf.catalog.plugin.OAuthPluginException;
 import ddf.catalog.source.UnsupportedQueryException;
 import java.io.IOException;
 import java.util.Date;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Function;
 import org.codice.ddf.catalog.ui.metacard.EntityTooLargeException;
 import org.codice.ddf.catalog.ui.query.cql.CqlRequestImpl;
+import org.codice.ddf.catalog.ui.query.cql.SourceWarningsFilterManager;
 import org.codice.ddf.catalog.ui.query.geofeature.FeatureService;
 import org.codice.ddf.catalog.ui.query.handlers.CqlTransformHandler;
 import org.codice.ddf.catalog.ui.query.suggestion.DmsCoordinateProcessor;
@@ -81,6 +84,8 @@ public class QueryApplication implements SparkApplication, Function {
 
   private final UtmUpsCoordinateProcessor utmUpsCoordinateProcessor;
 
+  private final SourceWarningsFilterManager sourceWarningsFilterManager;
+
   private FeatureService featureService;
 
   private CqlTransformHandler cqlTransformHandler;
@@ -97,13 +102,15 @@ public class QueryApplication implements SparkApplication, Function {
       LatLonCoordinateProcessor latLonCoordinateProcessor,
       DmsCoordinateProcessor dmsCoordinateProcessor,
       MgrsCoordinateProcessor mgrsCoordinateProcessor,
-      UtmUpsCoordinateProcessor utmUpsCoordinateProcessor) {
+      UtmUpsCoordinateProcessor utmUpsCoordinateProcessor,
+      SourceWarningsFilterManager sourceWarningsFilterManager) {
     this.latLonCoordinateProcessor = latLonCoordinateProcessor;
     this.dmsCoordinateProcessor = dmsCoordinateProcessor;
     this.mgrsCoordinateProcessor = mgrsCoordinateProcessor;
     this.utmUpsCoordinateProcessor = utmUpsCoordinateProcessor;
     this.cqlTransformHandler = cqlTransformHandler;
     this.cqlValidationHandler = cqlValidationHandler;
+    this.sourceWarningsFilterManager = sourceWarningsFilterManager;
   }
 
   @Override
@@ -117,6 +124,7 @@ public class QueryApplication implements SparkApplication, Function {
           try {
             CqlRequestImpl cqlRequest = GSON.fromJson(util.safeGetBody(req), CqlRequestImpl.class);
             CqlQueryResponse cqlQueryResponse = cqlQueryUtil.executeCqlQuery(cqlRequest);
+            addApplicableWarningsTo(cqlQueryResponse);
             return GSON.toJson(cqlQueryResponse);
           } catch (OAuthPluginException e) {
             res.status(e.getErrorType().getStatusCode());
@@ -233,5 +241,30 @@ public class QueryApplication implements SparkApplication, Function {
 
   public void setCqlQueryUtil(CqlQueriesImpl cqlQueryUtil) {
     this.cqlQueryUtil = cqlQueryUtil;
+  }
+
+  private void addApplicableWarningsTo(CqlQueryResponse response) {
+    if (response.getQueryResponse() == null) {
+      LOGGER.debug(
+          "could not add warnings to the response {} because the field of the response which would contain the warnings is null",
+          response);
+      return;
+    }
+
+    if (response.getWarnings() == null) {
+      LOGGER.debug(
+          "could not add warnings to the response {} because the application received a null value when it got the response's warnings",
+          response);
+      return;
+    }
+
+    response
+        .getQueryResponse()
+        .getProcessingDetails()
+        .stream()
+        .map(sourceWarningsFilterManager::filterWarningsOf)
+        .map(ProcessingDetails::getWarnings)
+        .filter(Objects::nonNull)
+        .forEach(response.getWarnings()::addAll);
   }
 }
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlQueryResponseImpl.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlQueryResponseImpl.java
@@ -32,6 +32,7 @@ import ddf.catalog.source.UnsupportedQueryException;
 import ddf.catalog.source.solr.SolrMetacardClientImpl;
 import java.io.Serializable;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -71,6 +72,8 @@ public class CqlQueryResponseImpl implements CqlQueryResponse {
   private final Boolean userSpellcheckIsOn;
 
   private final List<ResultHighlight> highlights;
+
+  private final Set<String> warnings = new HashSet<>();
 
   // Transient so as not to be serialized to/from JSON
   private final transient QueryResponse queryResponse;
@@ -205,5 +208,9 @@ public class CqlQueryResponseImpl implements CqlQueryResponse {
 
   public Status getStatus() {
     return status;
+  }
+
+  public Set<String> getWarnings() {
+    return warnings;
   }
 }
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManager.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManager.java
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.query.cql;
+
+import ddf.catalog.operation.ProcessingDetails;
+import ddf.catalog.operation.impl.ProcessingDetailsImpl;
+import ddf.catalog.security.SourceWarningsFilter;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * A {@code SourceWarningsFilterManager} {@linkplain #filterWarningsOf(ProcessingDetails) filters
+ * the {@code warnings} of} {@link ProcessingDetails} with the {@linkplain
+ * #getBestCompatibleFilterFor(ProcessingDetails) best compatible} {@link SourceWarningsFilter} in
+ * its {@code filters}} such that the {@link ProcessingDetails} only contain {@code warnings} which
+ * the {@link SourceWarningsFilter} deems useful to the user.
+ */
+public class SourceWarningsFilterManager {
+
+  private final List<SourceWarningsFilter> filters;
+
+  public SourceWarningsFilterManager(List<SourceWarningsFilter> filters) {
+    if (filters == null) {
+      throw new IllegalArgumentException(
+          "the constructor of SourceWarningsFilterManager received a null List of Filters");
+    }
+
+    this.filters = filters;
+  }
+
+  public ProcessingDetails filterWarningsOf(ProcessingDetails processingDetails) {
+    if (!canInspect(processingDetails)) {
+      return new ProcessingDetailsImpl();
+    }
+
+    return getBestCompatibleFilterFor(processingDetails)
+        .map(bestFilter -> bestFilter.filter(processingDetails))
+        .orElse(new ProcessingDetailsImpl());
+  }
+
+  private boolean canInspect(ProcessingDetails processingDetails) {
+    return processingDetails != null && !filters.isEmpty();
+  }
+
+  private Optional<SourceWarningsFilter> getBestCompatibleFilterFor(
+      ProcessingDetails processingDetails) {
+    return filters
+        .stream()
+        .filter(Objects::nonNull)
+        .filter(filter -> filter.canFilter(processingDetails))
+        .findFirst();
+  }
+}
--- a/ui-backend/catalog-ui-search/src/test/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManagerTest.java
+++ b/ui-backend/catalog-ui-search/src/test/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManagerTest.java
@@ -1,0 +1,191 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.query.cql;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import ddf.catalog.operation.ProcessingDetails;
+import ddf.catalog.security.SourceWarningsFilter;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.junit.Test;
+
+public class SourceWarningsFilterManagerTest {
+
+  @Test
+  public void testNullProcessingDetailsProduceProcessingDetailsWithEmptyWarnings() {
+    SourceWarningsFilter mockedFilter = mock(SourceWarningsFilter.class);
+    when(mockedFilter.canFilter(any())).thenReturn(true);
+
+    ProcessingDetails mockedDetailsWithNonEmptyWarnings = mock(ProcessingDetails.class);
+    when(mockedDetailsWithNonEmptyWarnings.getWarnings())
+        .thenReturn(Collections.singletonList("doesn't look empty to me, boss"));
+    when(mockedFilter.filter(any())).thenReturn(mockedDetailsWithNonEmptyWarnings);
+
+    SourceWarningsFilterManager filterManager =
+        new SourceWarningsFilterManager(Collections.singletonList(mockedFilter));
+
+    ProcessingDetails detailsWithFilteredWarnings = filterManager.filterWarningsOf(null);
+
+    assertThat(detailsWithFilteredWarnings, is(notNullValue()));
+
+    assertThat(detailsWithFilteredWarnings.getWarnings(), is(Collections.emptyList()));
+  }
+
+  @Test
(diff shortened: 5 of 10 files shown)


kentmorrissey (commented):
I think that I should test the changes which this commit makes to `QueryApplication.java`, but I don't know how I would. Thoughts?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlQueryResponseImpl.java
```diff
@@ -198,4 +201,11 @@ public String getId() {
   public Status getStatus() {
     return status;
   }
+
+  public void addToWarnings(List<String> unaddedWarnings) {
+    if (unaddedWarnings != null && !unaddedWarnings.isEmpty()) {
+      warnings.addAll(
+          unaddedWarnings.stream().filter(Objects::nonNull).collect(Collectors.toSet()));
+    }
+  }
```
kentmorrissey: Should I write tests for this method? It doesn't do much, but tests for when `addedWarnings` takes a null or empty value would provide value. I would have added those tests already, except that I couldn't find `CqlQueryResponseImplTest.java`, and I don't know that this method justifies the creation of `CqlQueryResponseImplTest.java` for maybe three tests which wouldn't provide much value.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -97,13 +102,15 @@ public QueryApplication(
...
-      UtmUpsCoordinateProcessor utmUpsCoordinateProcessor) {
+      UtmUpsCoordinateProcessor utmUpsCoordinateProcessor,
+      SourceWarningsFilterManager sourceWarningsFilterManager) {
     this.latLonCoordinateProcessor = latLonCoordinateProcessor;
     this.dmsCoordinateProcessor = dmsCoordinateProcessor;
     this.mgrsCoordinateProcessor = mgrsCoordinateProcessor;
     this.utmUpsCoordinateProcessor = utmUpsCoordinateProcessor;
     this.cqlTransformHandler = cqlTransformHandler;
     this.cqlValidationHandler = cqlValidationHandler;
+    this.sourceWarningsFilterManager = sourceWarningsFilterManager;
```
Lambeaux, line 113: 📖 Typically we would check for `null` in the constructor so we don't have to check on every call to the object in question. Since none of the other objects are being checked, we are deferring the check of this to blueprint by specifying that any looked up services are "mandatory". If we don't specify that in the blueprint, I believe it is the default.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -117,6 +124,17 @@ public void init() {
...
+            if (cqlQueryResponse != null
+                && (queryResponse = cqlQueryResponse.getQueryResponse()) != null
+                && sourceWarningsFilterManager != null) {
+              queryResponse
+                  .getProcessingDetails()
+                  .stream()
+                  .map(sourceWarningsFilterManager::filterWarningsOf)
+                  .map(ProcessingDetails::getWarnings)
+                  .forEach(cqlQueryResponse::addToWarnings);
+            }
```
Lambeaux: ❗ Don't need the `sourceWarningsFilterManager != null` check, and if we did, we should be handling that in the constructor so that those kinds of errors are detected during wiring / startup instead of on the first request.

Lambeaux: ❓ This is really hard to read. How about this instead? 
```
if (cqlQueryResponse != null) {
  Optional.ofNullable(cqlQueryResponse.getQueryResponse())
    .map(CqlQueryResponse::getProcessingDetails)
    .flatMap(Collection::stream)
    .map(sourceWarningsFilterManager::filterWarningsOf)
    .map(ProcessingDetails::getWarnings)
    .forEach(cqlQueryResponse::addToWarnings);
}
```

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlQueryResponseImpl.java
Lambeaux: That's fine. As long as we test the code that is producing the warnings to ensure they're not `null` it should be fine 👍 According to a hashset `null` is a valid unique value compared to other objects meaning a set either contains `null` or it doesn't. So the real value comes from testing the objects that might produce `null` and sending it to this set. 

Another option is to change this function to `throw` an `IllegalArgumentException` if `null` is contained within the input set. That's an easy condition to write one unit test for.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManager.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * A {@code SourceWarningsFilterManager} {@linkplain #filterWarningsOf(ProcessingDetails) filters
+ * the {@code warnings} of} {@link ProcessingDetails} with the {@linkplain
+ * #getBestCompatibleFilterFor(ProcessingDetails) best compatible} {@link SourceWarningsFilter} in
+ * its {@code filters}} such that the {@link ProcessingDetails} only contain {@code warnings} which
+ * the {@link SourceWarningsFilter} deems useful to the user.
+ */
+public class SourceWarningsFilterManager {
+
+  private List<SourceWarningsFilter> filters;
```
Lambeaux: ✏️ this can be `final`

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManager.java
```diff
@@ -0,0 +1,68 @@
...
+  public ProcessingDetails filterWarningsOf(ProcessingDetails processingDetails) {
+    SourceWarningsFilter bestCompatibleFilter = null;
+
+    if (canInspect(processingDetails)) {
+      bestCompatibleFilter = getBestCompatibleFilterFor(processingDetails);
+    }
+
+    return bestCompatibleFilter == null
+        ? new ProcessingDetailsImpl()
+        : bestCompatibleFilter.filter(processingDetails);
```
Lambeaux: ❗ Instead of maintaining intermediate state, `return` early. It'll cut down on the code. It also helps keep branching complexity low. 
```
if (canInspect(processingDetails)) {
  return getBestCompatibleFilterFor(processingDetails).filter(processingDetails);
}
return new ProcessingDetailsImpl();
```
This was my first instinct. But, this doesn't work because `getBestCompatibleFilterFor` is  `@Nullable`. What do you think of this instead? Notice the negated predicate. 
```
if (!canInspect(processingDetails)) {
  return new ProcessingDetailsImpl();
}
return Optional.ofNullable(getBestCompatibleFilterFor(processingDetails))
  .map(f -> f.filter(processingDetails))
  .orElse(new ProcessingDetailsImpl());
```

ui-backend/catalog-ui-search/src/test/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManagerTest.java
```diff
@@ -0,0 +1,191 @@
...
+import static org.mockito.Mockito.when;
+
+import ddf.catalog.operation.ProcessingDetails;
+import ddf.catalog.security.SourceWarningsFilter;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.junit.Test;
+
+public class SourceWarningsFilterManagerTest {
```
Lambeaux, line 30: ✏️ If you want more concise tests, annotate the test class with `@RunWith(MockitoJUnitRunner.class)` which will allow you to declare class-level mocks that get reset after every test, like so: 
```
@RunWith(MockitoJUnitRunner.class)
public class SourceWarningsFilterManagerTest {

@Mock private SourceWarningsFilter mockedFilter;
@Mock private ProcessingDetails mockedDetails;

...
```
If you find yourself recreating a lot of the same mocks using the `mock(<class>)` function, this can help.

ui-backend/catalog-ui-search/src/test/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManagerTest.java
```diff
@@ -0,0 +1,191 @@
...
+    assertThat(detailsWithFilteredWarnings.getWarnings(), is(Collections.emptyList()));
+  }
+
+  @Test
+  public void testEmptyFiltersProduceProcessingDetailsWithEmptyWarnings() {
+    SourceWarningsFilterManager filterManager =
+        new SourceWarningsFilterManager(Collections.emptyList());
+
+    ProcessingDetails mockedDetailsWithSourceId = mock(ProcessingDetails.class);
+    when(mockedDetailsWithSourceId.getSourceId()).thenReturn("source id");
```
Lambeaux, line 58: ✏️ Testing constants can be moved up higher in scope like so:
```
public class SourceWarningsFilterManagerTest {

private static final String FAKE_SOURCE_ID = "source id";
private static final String FAKE_OTHER_THING = "...";

...
```
Note the common prefix of `FAKE_` to benefit from simpler auto-complete.

Lambeaux (changes requested):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlQueryResponseImpl.java
Lambeaux: Regarding your comment about the overhead of a new class, for production code I would try to avoid it, but for test code I think it's a moot point. At least, in this instance. There are always exceptions.

ui-backend/catalog-ui-search-api/src/main/java/org/codice/ddf/catalog/ui/query/utility/CqlQueryResponse.java
```diff
@@ -32,4 +32,6 @@
   String getId();
 
   Status getStatus();
+
+  void addToWarnings(List<String> unaddedWarnings);
```
brhumphe: ❓ What purpose does it serve to add warnings without providing a mechanism to retrieve warnings in this interface?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManager.java
```diff
@@ -0,0 +1,68 @@
...
+ * the {@link SourceWarningsFilter} deems useful to the user.
+ */
+public class SourceWarningsFilterManager {
+
+  private List<SourceWarningsFilter> filters;
+
+  public SourceWarningsFilterManager(List<SourceWarningsFilter> filters) {
+    if (filters == null) {
+      throw new IllegalArgumentException(
+          "the constructor of SourceWarningsFilterManager received a null argument");
```
brianfelix: ✏️ Since we know the filters were null, instead of saying "a null argument", consider saying "a null list of filters" or something similar.

kentmorrissey:
build now

kentmorrissey (commented):
Ignore this comment.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -117,6 +124,16 @@ public void init() {
           try {
             CqlRequestImpl cqlRequest = GSON.fromJson(util.safeGetBody(req), CqlRequestImpl.class);
             CqlQueryResponse cqlQueryResponse = cqlQueryUtil.executeCqlQuery(cqlRequest);
+            if (cqlQueryResponse.getQueryResponse() != null) {
+              cqlQueryResponse
+                  .getQueryResponse()
+                  .getProcessingDetails()
+                  .stream()
+                  .map(sourceWarningsFilterManager::filterWarningsOf)
+                  .map(ProcessingDetails::getWarnings)
```
kentmorrissey: Should I add `.filter(Objects::nonNull)` between these lines?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -117,6 +124,16 @@ public void init() {
...
+                  .filter(Objects::nonNull)
+                  .forEach(cqlQueryResponse.getWarnings()::addAll);
```
kentmorrissey: Can we guarantee that each implementation of `CqlQueryResponse.getWarnings()` never returns null? I guess I could add a comment like `// must not return null` to the method and then make sure that each implementation has a test which verifies that `getWarnings()` never returns null. Alternatively, I could verify that `cqlQueryResponse.getWarnings()` doesn't return null before I call it here. Thoughts?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
Lambeaux: Between? No, not necessary. The `filterWarningsOf` method is not `@Nullable` and returns a blank impl when it needs to.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
Lambeaux: Probably a good idea to pull this entire `IF`/streaming block into its own method like so, which would reduce your nesting - note I've added to the `null` check but I've flipped the condition again:
```
private static void addSourceWarningsIfNecessary(CqlQueryResponse response) {
  if (response.getQueryResponse == null || response.getWarnings == null) {
    return;
  }
  response.getQueryResponse()
    .getProcessingDetails()
    .stream()
    // ... etc etc
    .filter(Objects::nonNull)
    .forEach(response.getWarnings()::addAll);
}
```
Then the calling code becomes:
```
CqlQueryResponse cqlQueryResponse = cqlQueryUtil.executeCqlQuery(cqlRequest);
addSourceWarningsIfNecessary(cqlQueryResponse);
```

Lambeaux (approved):
Looks good pending final comments.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
kentmorrissey: I've noted that you flipped the condition. Do you think that what you've suggested provides more clarity than
```
if (response.getQueryResponse != null && response.getWarnings != null) {
    response.getQueryResponse()
      .getProcessingDetails()
      .stream()
      // ... etc
      .filter(Objects::nonNull)
      .forEach(response.getWarnings()::addAll);
}
```
within the private method?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/SourceWarningsFilterManager.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  @Nullable
+  private SourceWarningsFilter getBestCompatibleFilterFor(ProcessingDetails processingDetails) {
+    return filters
+        .stream()
+        .filter(Objects::nonNull)
+        .filter(filter -> filter.canFilter(processingDetails))
+        .findFirst()
+        .orElse(null);
```
mojogitoverhere: ✏️ If you remove `orElse(null)` and change the signature to `Optional<SourceWarningsFilter>` / remove the `@Nullable` annotation,  then `filterWarningsOf` can be simplified a bit:
```
  public ProcessingDetails filterWarningsOf(ProcessingDetails processingDetails) {
    if (!canInspect(processingDetails)) {
      return new ProcessingDetailsImpl();
    }

    return getBestCompatibleFilterFor(processingDetails) // <-- removed the optional wrapper here
        .map(bestFilter -> bestFilter.filter(processingDetails))
        .orElse(new ProcessingDetailsImpl());
  }

```

mojogitoverhere (approved):
LGTM after @Lambeaux 's comments are addressed

mojogitoverhere (approved):

mojogitoverhere:
build now

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
Lambeaux: @kentmorrissey Yes but in this particular case it's not obvious. I prefer structuring my `IF`s by consistently detecting bad cases and handling those eagerly. It'll make for a consistent codebase because it works for the trivial ☝️ and scales to the complex as well. 

A lot of it does come down to user preference, but it does empirically cut down on cyclomatic complexity of code. That's a win IMO. 

And if it doesn't - then I'd suggest doing what's necessary to cut down cyclomatic complexity unless there's a higher priority given the context.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -234,4 +242,19 @@ public void setEndpointUtil(EndpointUtil util) {
   public void setCqlQueryUtil(CqlQueriesImpl cqlQueryUtil) {
     this.cqlQueryUtil = cqlQueryUtil;
   }
+
+  private void addApplicableWarningsTo(CqlQueryResponse response) {
+    if (response.getQueryResponse() == null || response.getWarnings() == null) {
+      return;
```
Lambeaux, line 258: ❓ Should we `debug` log here?

Lambeaux (approved):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
kentmorrissey: I think so. Good catch.

brianfelix (approved):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
Lambeaux: ❓ Less is more, and data speaks louder than words. How about this instead?
```
LOGGER.debug("Could not add warnings to response, the response was null: {}", response);
```
and
```
LOGGER.debug("Could not add warnings to response, list of warnings was null: {}", response);
```

It minimizes the number of renames that have to occur if we refactor. Also be sure to test that calling `toString()` on any `CqlQueryResponse` object yields a useful object string, with all nested objects usefully written out as well. 

Typically when we log debug info we should also log relevant data for those troubleshooting the system.

Lambeaux (approved):
When it comes to log messages, less is more, and nothing is better than seeing the problematic data firsthand.

mojogitoverhere (approved):

brianfelix (approved):

mojogitoverhere:
## 🎉 Hero was successful! 🎉 ##
1. Applied and built patch from the test instructions
2. Setup a new OpenSearch source
3. Ran a query
4. Verified an announcement popped up with the OpenSearch source name as the title with the message `looks like the test passed!`

PR #193, opened 2020-05-04
DDF 5260 Updates editing restrictions to take into account write access #5895

jlcsmith:
**What does this PR do?**
Updates the edit restriction checking to check the write access for the result.

Removes the editingEnabled UI property as edit restrictions are determined based on metacard attributes.

Updates the usage of these restrictions to be local and removes the global 'is-editing-restricted' class that was placed on the body. This lets the choice be decided on a per result basis and makes it so components who want to know are in charge of figuring it out.

**Who is reviewing it?**
@andrewkfiedler
@cassandrabailey293 
@Bdthomson
@bdeining

Select relevant component teams:
@codice/ui

Ask 2 committers to review/merge the PR and tag them here.
@andrewkfiedler
@Bdthomson
@bdeining


ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/User.js
```diff
@@ -499,11 +499,29 @@ User.Response = Backbone.AssociatedModel.extend({
...
-      metacard.get('createdBy') !== 'system'
-    )
+  /**
+   * Allow workspaces to be created no matter what,
+   * but still enforce the other restrictions (if it's a shared workspace,
+   * check if they have permission to write to it)
+   *
+   * For all other types, every policy is enforced.
+   */
+  canWrite(thing) {
```
cassandrabailey293, line 509: Looks like this `isEditingRestricted` is set by a backend property `isEditingAllowed`. I think for the result forms purpose, we will leverage this property and it should correctly return false here. However, for the converse case (when the result form is editable), the logic is broken for result forms. They currently don't have a `type` property, so the `default` case below will be hit, causing a console error.

It seems like the `default` case below is assuming that there will be a collection of some sort if it is hit, is that correct?

cassandrabailey293: the added check for `undefined` in the `default` case below handles the result forms, so no addiitonal changes should be needed for them. thanks!

jlcsmith: @cassandrabailey293 - made some updates to this after speaking with @andrewkfiedler .  This logic now appears to work as expected in DDF.  It would need verification for downstream projects though.  Note that I removed the editingEnabled property as I don't think it is needed since this change determines editability based on the metacard's attributes.

andrewkfiedler (approved):

cassandrabailey293:
Did some testing of ddf & downstream, looks like there is no regression. LGTM 👍

cassandrabailey293 (approved):

bdeining (approved):

PR #209, opened 2020-05-06
DDF-6008 Return query highlighting results to the UI

derekwilhelm:
This enables the highlighting results from a query to be returned to the UI.

To test:
1) build this DDF PR: https://github.com/codice/ddf/pull/6009 
2) build this PR
3) Start SolrCloud
4) Modify the schema in the catalog collection to store the tokenized text fields (stored="true") 
5) Start DDF and do a standard profile install
6) Add ddf-ui front and back end
7) from the karaf console enable the highlighting capability: `property solr.highlight.enabled true`
8) Ingest data into DDF
9) perform a text query using specific words (i.e. not a wildcard query)
10) verify that the cql response contains the highlighting information.


criscrasbell (approved):

jlcsmith (approved):

vinamartin (approved):
Just tested this ✅ LGTM

vinamartin:
❓ I'm not sure if this is related to your PR or not, but I'm getting non-deterministic results back with the resultIds- sometimes I get the source ID prepended and sometimes I don't. 
My guess is this has something to do with caching, but I'm confused why that would be the case.

derekwilhelm:
The ID discrepancy had to do with the cache.  The cache modifies the ID to include the source so that is what Solr sees as the ID.  I updated my DDF PR to fix this so that the cache will return proper highlight ids, just like it does with the actual metacard IDs in the results.

PR #211, opened 2020-05-06
Updates SecurityTokenHolder name to PrincipalHolder

rzwiefel:
Updates according to a name change that happened in ddf master (2.24.0-SNAPSHOT) that will resolve the issue when 2.24.0 has been released.

Long term the plan is to move the SecureWebSocketServlet class to DDF.


jlcsmith:
Close to https://github.com/codice/ddf-ui/pull/209 ?

PR #213, opened 2020-05-12
DDF-5900 moved SubjectUtils, Permissions, and SecurityLogger to services

stustison:
@SmithJosh @garrettfreibott @blen-desta @bakejeyner @pklinef @rzwiefel

PR #249, opened 2020-06-10
Updates version ranges of pac4j and nimbus to work with ddf versions

rzwiefel:
## Verified 2020-06-10 against codice/ddf de9e1752f5

This has a[ commit from another PR](https://github.com/codice/ddf-ui/pull/213) which is also needed to make ddf-ui work with current codice/ddf/2.25.0-SNAPSHOT .. it will disappear when that pr is merged and this is rebased.

------
❗ This pr will need to be updated to remove `2.25.0-SNAPSHOT` and changed to 2.25.0 release before it is merged.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,7 +51,7 @@
     <maven.deploy.plugin.version>2.8.2</maven.deploy.plugin.version>
 
     <!--Backend properties-->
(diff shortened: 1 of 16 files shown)


vinamartin:
Merged in https://github.com/codice/ddf-ui/pull/282

PR #260, opened 2020-07-06
[3.5.x] DDF-UI-258 refactor cql endpoint code

### LINKED ISSUE #258: Refactor CQL endpoint to to support downstream projects.
Downstream projects need to create a thin layer over the existing CQL endpoint.

glenhein:
A downstream project needs to create a thin layer over the existing cql endpoint. Some refactoring was needed.

This PR extracts the logic that creates the QueryRequest object into a service that can be used by the downstream project. Also, in the case the cql fails to parse, an exception is thrown instead of sending a "halt" to the HTTP layer. The QueryApplication catches that exception and issues to the same "halt" in order to maintain the same behavior.

closes #258

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
@@ -124,7 +124,7 @@ public class QueryApplication implements SparkApplication, Function {
             res.status(e.getErrorType().getStatusCode());
             return GSON.toJson(ImmutableMap.of(ID_KEY, e.getSourceId(), URL_KEY, e.getUrl()));
           } catch (CqlParseException e) {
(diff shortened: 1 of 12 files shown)


ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/QueryApplication.java
```diff
@@ -121,6 +123,10 @@ public void init() {
           } catch (OAuthPluginException e) {
             res.status(e.getErrorType().getStatusCode());
             return GSON.toJson(ImmutableMap.of(ID_KEY, e.getSourceId(), URL_KEY, e.getUrl()));
+          } catch (CqlParseException e) {
+            LOGGER.debug("Unalbe to parse CQL", e);
```
vinamartin: ✏️  "Unable"

glenhein:
Closed in favor of https://github.com/codice/ddf-ui/pull/267

PR #267, opened 2020-07-07
DDF-UI-258 refactor cql endpoint code

### LINKED ISSUE #258: Refactor CQL endpoint to to support downstream projects.

glenhein:
A downstream project needs to create a thin layer over the existing cql endpoint. Some refactoring was needed.

This PR extracts the logic that creates the QueryRequest object into a service that can be used by the downstream project. Also, in the case the cql fails to parse, an exception is thrown instead of sending a "halt" to the HTTP layer. The QueryApplication catches that exception and issues to the same "halt" in order to maintain the same behavior.

closes #258

### DIFF
--- a/ui-backend/catalog-ui-search-api/src/main/java/org/codice/ddf/catalog/ui/query/utility/QueryRequestFactory.java
+++ b/ui-backend/catalog-ui-search-api/src/main/java/org/codice/ddf/catalog/ui/query/utility/QueryRequestFactory.java
@@ -20,6 +20,12 @@ import java.util.Set;
 import javax.annotation.Nullable;
 import org.codice.ddf.catalog.ui.CqlParseException;
 
+/**
+ * This is a factory that builds a QueryRequest object based on several query parameters.
+ *
+ * <p><i>This code is experimental. While this interface is functional and tested, it may change or
+ * * be removed in a future version of the library.</i>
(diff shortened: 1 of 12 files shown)


ui-backend/catalog-ui-search-api/src/main/java/org/codice/ddf/catalog/ui/query/utility/QueryRequestFactory.java
```diff
@@ -0,0 +1,44 @@
...
+package org.codice.ddf.catalog.ui.query.utility;
+
+import ddf.catalog.filter.FilterBuilder;
+import ddf.catalog.operation.QueryRequest;
+import java.util.List;
+import java.util.Set;
+import javax.annotation.Nullable;
+import org.codice.ddf.catalog.ui.CqlParseException;
+
+public interface QueryRequestFactory {
```
jlcsmith, line 29: javadocs?
mark this as an experimental interface perhaps?

jlcsmith (approved):

vinamartin (approved):

jlcsmith:
🎉 Hero Successful 🎉 
Verified that queries of various types did not regress

Issue #248, opened 2020-06-09
When shareable items get shared, user must refresh to see it


cassandrabailey293:
When something(workspaces, result forms, search forms, etc) is shared by User A to User B, User B must reload the whole webpage to see the newly shared item. The item should appear automatically.

PR #295, opened 2020-07-29
[3.4.x] DDF-UI-248 Update UI automatically on share/unshare/delete

cassandrabailey293:
Forward port of https://github.com/codice/ddf/pull/6143

____________________________________________________________________________________________________________

#### What does this PR do?
Add Server-Sent Event messaging to DDF, and uses it to automatically update search forms, workspaces, and result forms on share, unshare, and delete operations 

#### Who is reviewing it? 

@andrewzimmer @bdeining @hayleynorton @zta6 @abel-connexta 

@codice/ui 

#### Ask 2 committers to review/merge the PR and tag them here.

@pklinef 
@rymach 

### Setup
* build 
* before starting ddf go into /etc/ directory, update users.attributes file to include
```
"test" : {
       "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress":"test@localhost.local",
       "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/role": ["manager","viewer","systembundles"]
   }
```
and in users.properties files add
`test=test,group,manager,viewer,systembundles,ssh`

* start ddf, profile install, and start the dev server
* log into the test user on `localhost:8993/search/catalog`
* open another window and go to `localhost:8080/search/catalog`

### Share/Unshare
0. Navigate to the workspaces collection page on `admin` user's instance
1. Create a new workspace on `test` user's instance
2. Share the workspace with the `admin` group (should be there already)
3. Verify on the `admin` instance that `test` user's workspace shows up automatically (takes 3 seconds)
4. Share the other way around (from `admin` to `test@localhost.local`) and verify it shows up on `test`'s workspace collection automatically

Repeat these steps for unshare. The workspace should disappear from the user to whom the workspace was unshared.

### Delete
0-3. Do steps 0-3 from *Share/Unshare*
4. Delete the workspace on `test` user's page
5. Verify on the `admin` instance that `test` user's workspace disappears automatically (takes 3 seconds)
6. Delete the other way around and verify

Repeat these steps for search forms and result forms.

### Workspace Search
0. Open a workspace on `admin`'s instance and Click on the Search DDF button
1. Create a new search form on `test` user's instance
2. Share the search form with the `admin` group (should be there already)
3. Verify on the `admin` instance that `test` user's search form shows up in the three-dot menu drop down when you click "Use Other Search Form" (takes 3 seconds)
4. Create a new result form on the `test` user's instance
5. Share that result form with `admin`
6. Verify in `admin` user's instance that the result form shows up in the Result Form dropdown in the advanced search from.  

Fixes: https://github.com/codice/ddf-ui/issues/248

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:


ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+  private static final String HEARTBEAT = ": \n\n";
+
+  @Override
+  public void init() {
+    get(
+        "/events/:id",
+        (req, res) -> {
+          try {
+            String sourceId = req.params(":id");
+            if (listeners.size() >= 5) {
```
pklinef, line 46: ❗ I think this will block five event connections for all clients to this system, not per browser.  Try to simulate more than 5 browsers opening event connections to verify.

https://stackoverflow.com/questions/31238626/curl-structuring-request-to-validate-server-sent-events

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+import java.io.PrintWriter;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import spark.servlet.SparkApplication;
+
+public class EventApplication implements SparkApplication {
```
pklinef, line 27: ❓  Looks like Javalin (a fork of Spark) supports SSE.  Was there any investigation into switching to Javalin?  We have for other Spark endpoints downstream.

https://javalin.io/news/2019/01/17/javalin-2.6.0-released.html
https://javalin.io/documentation#server-sent-events

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+          } catch (Exception e) {
+            LOGGER.error("Event Source configuration error");
+          }
+          res.status(500);
+          return "Event Source configuration error: Server error";
+        });
+  }
+
+  public static void notifyListeners(EventType type) {
+    ExecutorService es = Executors.newSingleThreadExecutor();
```
pklinef, line 76: ❗ It looks like a lot of executors are being created and never shutdown.  Maybe we should consider using `Executors.newCachedThreadPool()` and set it to a field on this class.  Then all of the notify methods could use that pool.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executors.html#newCachedThreadPool--

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+          }
+        });
+    es.submit(
+        () -> {
+          synchronized (listeners) {
+            listeners
+                .values()
+                .forEach(
+                    (listener) -> {
+                      listener.write("event: " + type.getType() + "\n");
```
pklinef, line 92: ❗ If the `PrintWriter` throws any exceptions, all remaining listeners will not be notified.  Should add a try-catch.  The listener that failed might need to be removed if the exception is related to the `PrintWriter` being closed.

>Performs the given action for each element of the Iterable until all elements have been processed or the action throws an exception.
https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html#forEach-java.util.function.Consumer-

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import spark.servlet.SparkApplication;
+
+public class EventApplication implements SparkApplication {
+
+  private static final Map<String, PrintWriter> listeners = new ConcurrentHashMap<>();
```
pklinef, line 29: ❗ If there are any leaks, this map could slowly grow until the system OOMs.  We should consider using a timed eviction cache to avoid that.

https://github.com/google/guava/wiki/CachesExplained#timed-eviction

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+            while (!out.checkError()) {
+              out.write(HEARTBEAT);
+              out.flush();
+              Thread.sleep(15000);
+            }
+            notifyListener(sourceId, new EventType("CLOSE"));
+            listeners.remove(sourceId);
+            return "Event Source connection closed";
+          } catch (Exception e) {
+            LOGGER.error("Event Source configuration error");
```
pklinef, line 68: ❗ Listeners could be orphaned in the hash map if the remove listener is skipped.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+
+            res.status(200);
+            listeners.put(sourceId, out);
+            while (!out.checkError()) {
+              out.write(HEARTBEAT);
+              out.flush();
+              Thread.sleep(15000);
+            }
+            notifyListener(sourceId, new EventType("CLOSE"));
+            listeners.remove(sourceId);
```
pklinef, line 65: ❗ Some runtime exceptions could cause this line to be skipped which would orphan those listeners in the map.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+          return "Event Source configuration error: Server error";
+        });
+  }
+
+  public static void notifyListeners(EventType type) {
+    ExecutorService es = Executors.newSingleThreadExecutor();
+    es.submit(
+        () -> {
+          try {
+            Thread.sleep(3000);
```
pklinef, line 80: ❓ Why are these 3 second sleeps required?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+          } catch (InterruptedException e) {
+            LOGGER.error("Event Source notification error");
+          }
+        });
+    es.submit(
+        () -> {
+          synchronized (listeners) {
```
pklinef, line 87: ❓ Why are you locking on `listeners`?  It is a `ConcurrentHashMap` and you also do `put`s into it without synchronizing.

Are you doing this to try and avoid messages from interleaving?  If so, then I would suggest removing all `synchronized` blocks and instead make sure all listener writes are atomic by combining the two writes together instead.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -335,6 +340,17 @@ public void init() {
         (req, res) -> {
           String body = util.safeGetBody(req);
           List<MetacardChanges> metacardChanges = GSON.fromJson(body, METACARD_CHANGES_LIST_TYPE);
+          metacardChanges.forEach(
```
pklinef, line 343: ❗ I am concerned how this will scale up to thousands or tens of thousands of concurrent users.  All result forms, search forms, workspaces, or any metacard change will be notified back to all browsers listening even if the users cannot see the given records due to sharing settings or is not part of the data currently on the client side.  Given the current locking in the `EventApplication` and lack of back pressure, I would be worried that notifications could fall behind and start to backup.  If it does not catch up in time, the system could OOM or fail in other ways.

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Workspace.collection.js
```diff
@@ -40,6 +42,12 @@ module.exports = Backbone.Collection.extend({
       })
     })
     this.listenTo(this, 'add', this.tagGuestWorkspace)
+    const onMessage = () => {
+      this.fetch({ remove: true, merge: false })
```
pklinef, line 46: ❗ The current events do not include the ID of the added, updated, or deleted record.  This forces all clients getting the message to request all documents again which puts a lot more strain on the system when there are a large number of concurrent users.  If we sent the ID back, we could remove deleted workspaces on the client side with any calls.  We could request just the added or updated workspaces instead of all workspaces to reduce data transfer.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
```diff
@@ -0,0 +1,142 @@
...
+  // HEARTBEAT is used to write silent messages to keep the client connection. It is also used to
+  // trigger events
+  // by sending it with the data: tag. For more information, see
+  // https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Event_stream_format
+  private static final String HEARTBEAT = ": \n\n";
+
+  @Override
+  public void init() {
+    get(
+        "/events/:id",
```
pklinef, line 42: ❗ I don't think this is going to work when the system is horizontally scaled behind a load balancer.  For example, a client could connect to the event endpoint on system A.  Another client could connect to system B.  Events that happen on A will never be notified to listeners on B and vice versa.  Currently DDF UI endpoints are stateless except for state that is propagated back to the provider which is generally Solr which is shared by all systems.  Events would need to be propagated to all nodes by either calling them directly or using a shared message broker like Artemis or Kafka to be shared between all systems to all listeners could be notified.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-form/result-form.collection.js
```diff
@@ -68,6 +77,17 @@ module.exports = Backbone.AssociatedModel.extend({
   ],
   addResultForms() {
     if (!this.isDestroyed) {
+      const formsToDelete = this.get('resultForms')
```
pklinef, line 80: ✏️ This could be done directly in the event listener if the events also included an ID and if it was added, updated, or deleted.  We could modify the collection directly if it is a delete without a round trip to the server.  We could also just ask for the relevant added/updated records if they are present in the current collection.

pklinef (changes requested):

ui-backend/catalog-ui-search-api/src/main/java/org/codice/ddf/catalog/ui/events/EventTypeDeserializer.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.events;
+
+import com.google.gson.*;
```
rzwiefel, line 16: ❗ `*` Imports should be removed in favor of explicit exports. I know sometimes your IDE will sneak up and do this unawares, there should be a setting for it in intellij in case you don't have it set to stop it.
image

ui-backend/catalog-ui-search/pom.xml
```diff
@@ -471,6 +476,7 @@
                         <Export-Package>
                             org.codice.ddf.catalog.ui.metacard.internal,
                             org.codice.ddf.catalog.ui.alias,
+                            org.codice.ddf.catalog.ui.events
```
rzwiefel, line 479: ❓ It looks like the types you've defined for that package are in the `catalog-ui-search-api` module, did you mean to add it to the bundle-plugin definition on that pom instead of the `catalog-ui-search` pom?

pklinef:
Is there a plan to port this to master too?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
rymach: this is a deal breaker, right?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -335,6 +340,17 @@ public void init() {
...
+              change -> {
+                EventType type = change.getType();
+                if (type != null) {
+                  EventApplication.notifyListeners(type);
+                } else {
+                  String unknownType =
+                      new Gson().fromJson(body, JsonObject.class).get("type").getAsString();
```
rzwiefel, line 350: ❓ If `body` is the same `List<MetacardChanges>` that we already parsed, it doesn't seem like there would be a `type` property since it is a list of `MetacardChanges`. 

If you want the unknown type to be the default you can always add logic in the `getType` method to return `unknown` if it is null. If you do set it to return unknown type by default, then this whole block  could just become `change -> EventApplication.notifyListeners(change.getType())`

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
pklinef: Due to some of our AWS and other cloud deployments, I think it is a blocker for this PR in its current state unless I am missing something.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
cassandrabailey293: how would i simulate multiple browsers? Just send the request to open a connection more than 5 times?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
cassandrabailey293: I didn't look into it, but it looks like I could switch it over to Javalin if that's preferred.

ui-backend/catalog-ui-search-api/src/main/java/org/codice/ddf/catalog/ui/events/EventTypeDeserializer.java
cassandrabailey293: thank you!

ui-backend/catalog-ui-search/pom.xml
cassandrabailey293: ~~Yes, I did, whoops!~~ Actually, i think i need this line in order to use `EventApplication` inside of GSR

cassandrabailey293:
> Is there a plan to port this to master too?

If i remember correctly, we asked @bellcc if other downstream projects might use it and he said unlikely, so I was planning on not opening a PR against master.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
cassandrabailey293: is there a way to test this? when i was initially doing my research I saw that some people had SSE set up with AWS load balancers

rymach: even if we assume server sent events work fine through the load balancer, this feature still won't work unless the two servers are communicating somehow. if `server A jvm ddf instance` makes a workspace change there's no way for `server B jvm ddf instance` to know about it

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
pklinef: You can use that curl example in multiple terminal windows to simulate more than 5 browsers running in concurrently locally.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
pklinef: Javalin is the direction we would like to go since it supports using it as a servlet out of the box unlike Spark.  It is also nice to use their implementation since it has probably be more widely used and tested than our current version.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/events/EventApplication.java
pklinef: There are three different kinds of AWS load balancers.  I wonder if SSE work with them all.  I could not find anything definitive.

cassandrabailey293:
Closing due to https://github.com/codice/ddf-ui/pull/295#discussion_r463198297

PR #297, opened 2020-08-03
Forward port of 2.20x Performance Enhancements

andrewkfiedler:


andrewkfiedler:
This probably won't work at the moment, there was a recent change to CqlRequestImpl with a merge conflict that I couldn't figure out at the moment.  If anyone wants to take a crack at it, the premerge change is here:
https://github.com/andrewkfiedler/ddf-ui/blob/2.20.x-ForwardPort/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlRequestImpl.java#L234
 and the new file is:
https://github.com/codice/ddf-ui/blob/master/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/CqlRequestImpl.java

So in this PR, the lines from 178 to 200 in CqlRequestImpl need to be updated (removed, or moved?) before this will actually work.

jlcsmith:
It appears that that logic in CqlRequestImpl was moved to https://github.com/codice/ddf-ui/blob/master/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/cql/QueryRequestBuilder.java#L177-L191

ui-frontend/lerna-debug.log
```diff
@@ -0,0 +1,10 @@
+0 silly argv { _: [ 'run' ],
```
willwill96, line 1: Should this be moved to the gitignore?

andrewkfiedler: Yeah, not sure how it snuck in 😂

PR #306, opened 2020-08-05
Updates patch endpoint to optionally take source

vinamartin:
Made with help from the man, the myth, the legend: @rzwiefel 
This updates the patch endpoint to optionally accept the source just in case it is required for updating (example: harvested sources).

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
@@ -88,7 +88,9 @@ import java.io.OutputStream;
 import java.io.Serializable;
 import java.lang.reflect.Type;
 import java.time.Instant;
+import java.util.AbstractMap;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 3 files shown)


ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/EndpointUtil.java
```diff
@@ -308,6 +309,26 @@ public String getResponseWrapper(String responseType, Object response) {
...
+
+  public Map<String, Result> getMetacardsWithTagByAttributes(
+      String attributeName,
+      Collection<String> attributeValues,
+      String tag,
+      @Nullable Set<String> sourceIds) {
+    return getMetacardsWithTagByAttributes(
+        attributeName,
+        attributeValues,
+        filterBuilder.attribute(Core.METACARD_TAGS).is().like().text(tag),
```
vinamartin, line 323: 📝  Gonna be honest- I just added this in to make things easier to test because mocking filter builder is a pain

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -339,8 +341,14 @@ public void init() {
         (req, res) -> {
           String body = util.safeGetBody(req);
           List<MetacardChanges> metacardChanges = GSON.fromJson(body, METACARD_CHANGES_LIST_TYPE);
-
-          UpdateResponse updateResponse = patchMetacards(metacardChanges, getSubjectIdentifier());
+          Set<String> sources =
+              Optional.ofNullable(req.queryParams("source"))
```
pklinef, line 345: ❗ Sources are read only.  But you could change this to support `CatalogStore` IDs instead which can be updated.

pklinef (changes requested):

PR #312, opened 2020-08-06
Updates to fix some bugs, updates for downstream extensions

andrewkfiedler:
* Removes all of handlebars custom helpers, fixes up some downstream things

* Update within nav to be more performant

* Fix type issues, allow no icon for route in nav

* Refactor settings to allow passing in custom settings downstream

* Updates path matching to work better, fixes bug in masked-text-field


andrewkfiedler:
3.6.x was merged into master so OBE

PR #336, opened 2020-08-20
[3.6.x] Fixes styling issues with System Usage Modal

vinamartin:
Rewrites System Usage Modal to make it consistent with theme
Screen Shot 2020-08-20 at 15 57 25

To test: 
Configure `ddf.platform.ui.config` System Usage title/Body, use html for the body if you like.
Verify system usage modal appears as expected. **Verify you can't click outside the modal to close.**
**Note**: There is a weird visual effect when loading in the page- I don't think this is a big deal but it is a little strange.


emmberk:
I heroed this in a downstream project. I also confirmed that the model appears every time I refresh the page, as expected. Thanks for the quick fix, @vinamartin!

Note: I'm approving, but I didn't actually review the code changes.

emmberk (approved):

PR #352, opened 2020-09-01
[3.6.x] DDF-351 Extension Point for Security Logging

cantstoptheunk:
- Adds a security logging function that can be used downstream.
- If a downstream application does not fill out a function definition then the code will no-op and do nothing.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/result-selector.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/result-selector.tsx
@@ -31,10 +31,6 @@ import { Elevations } from '../theme/theme'
 import useTheme from '@material-ui/core/styles/useTheme'
 import SelectionRipple from '../golden-layout/selection-ripple'
 import { ResultType } from '../../js/model/Types'
-import Extensions from '../../extension-points'
-
-const postAuditLog = Extensions.postAuditLog
-let selectedIds = new Set<string>()
 
 const SelectedResults = ({ selectionInterface }: any) => {
   const selectedResults = useLazyResultsSelectedResultsFromSelectionInterface({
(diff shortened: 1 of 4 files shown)


andrewkfiedler (changes requested):
If this is centered around the inspector visual, shouldn't we be tying it into that visual rather than the result selector?

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/result-selector.tsx
```diff
@@ -31,7 +31,6 @@ import { Elevations } from '../theme/theme'
 import useTheme from '@material-ui/core/styles/useTheme'
 import SelectionRipple from '../golden-layout/selection-ripple'
 import { ResultType } from '../../js/model/Types'
-import Extensions from '../../extension-points'
```
cantstoptheunk: should probably not delete

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/inspector/inspector-lazy.tsx
```diff
@@ -22,14 +26,49 @@ const LazyInspector = ({ selectionInterface }: Props) => {
   React.useEffect(() => {
     selectionInterface.setSelectedResults(backboneModels)
   })
+  const conductAudit = () => {
+    let newSelectedIds = new Set<string>()
+    for (let key in selectedResults) {
```
cantstoptheunk: call it id instead of key

cantstoptheunk: object.keys() to grab the list of keys

criscrasbell (changes requested):
This extension point is too specific. It will only have one purpose. I think we should put the audit logging code entirely in DDF-UI.

andrewkfiedler:
Going to go a different route on this

PR #362, opened 2020-09-10
[3.6.x] Fixes displaying Archive/Overwrite tabs

vinamartin:
Fixes issue with displaying archive/overwrite tabs on results users didn't have write access to.
**Note**: ~I noticed this causes a pretty severe performance degradation on loading multiple results, looking into adding a loading state to lessen the blow here but I'm open to thoughts on how to make this check faster.~ This performance issue exists without this change.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/User.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/User.js
@@ -499,7 +499,7 @@ User.Response = Backbone.AssociatedModel.extend({
         return this.canWrite(thing.get('metacard').get('properties'))
       case 'query-result.collection':
       default:
-        if (this.some !== undefined) {
+        if (thing.some !== undefined) {
           !thing.some(subthing => {
             return !this.canWrite(subthing)
           })
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/tabs-metacard.view.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/tabs-metacard.view.js
@@ -130,5 +130,9 @@ module.exports = TabsView.extend({
       'is-preview-disabled',
       !properties.isMetacardPreviewEnabled()
     )
+    this.$el.toggleClass(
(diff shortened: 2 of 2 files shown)


jlcsmith (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/User.js
```diff
@@ -499,7 +499,7 @@ User.Response = Backbone.AssociatedModel.extend({
         return this.canWrite(thing.get('metacard').get('properties'))
       case 'query-result.collection':
       default:
-        if (this.some !== undefined) {
+        if (thing.some !== undefined) {
           !thing.some(subthing => {
```
vinamartin, line 503: 📝  I think the preferred solution here would be to exit out of this as soon as one result isn't editable. 
It would still be slow in the case all are editable, but it should lessen the load time for mixed results or when all aren't editable.

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/User.js
```diff
@@ -499,7 +499,7 @@ User.Response = Backbone.AssociatedModel.extend({
         return this.canWrite(thing.get('metacard').get('properties'))
       case 'query-result.collection':
       default:
-        if (this.some !== undefined) {
+        if (thing.some !== undefined) {
```
jordanwilking, line 502: 🤦 😂

jordanwilking (approved):

astagg (approved):

jordanwilking:
📑 Hero Successful 📑  

#### Steps

1. Built downstream and copied pr files into node_modules
1. As admin, ingested a resource and gave read access to user2
1. As user2, opened the inspector view for the resource. As expected, the archive/overwrite tabs were not visible.
image
1. As admin, elevated user2 privilege to write. As expected, tabs were then visible.
image
1. Verified that clicking on the tabs showed the correct views.

#### Notes
- It tooks a bit longer than usual for the tabs to update after I updated the permissions.  At first, it looked like elevating the permissions wasn't working, but after searching again a minute later it was working. Probably a soft commit issue.

PR #364, opened 2020-09-11
[3.6.x] DDF-363 Added fix to history endpoint to accept sources

cantstoptheunk:
Screen Shot 2020-09-11 at 12 42 10 PM

Build `DDF 2.25.1` and then this PR and then run `profile:install standard` when unzipping the distribution.
Then follow the instructions on https://github.com/codice/ddf-ui/blob/master/README.md (Ignore the optional steps).

Then upload a resource and edit some attributes and make sure that the history tab still creates history nodes.

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
@@ -142,7 +142,6 @@ import org.opengis.filter.Filter;
 import org.opengis.filter.sort.SortBy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import spark.QueryParamsMap;
 import spark.servlet.SparkApplication;
 
 public class MetacardApplication implements SparkApplication {
@@ -369,14 +368,10 @@ public class MetacardApplication implements SparkApplication {
         });
 
     get(
-        "/history/:id",
+        "/history/:id/:storeId",
         (req, res) -> {
           String id = req.params(":id");
-          QueryParamsMap paramsMap = req.queryMap().get("storeId");
-          String storeId = null;
-          if (paramsMap.hasValue()) {
-            storeId = paramsMap.value();
-          }
+          String storeId = req.params(":storeId");
           List<Result> queryResponse = getMetacardHistory(id, storeId);
           if (queryResponse.isEmpty()) {
             res.status(204);
@@ -398,15 +393,11 @@ public class MetacardApplication implements SparkApplication {
         });
 
     get(
-        "/history/revert/:id/:revertid",
+        "/history/revert/:id/:revertid/:storeId",
         (req, res) -> {
           String id = req.params(":id");
           String revertId = req.params(":revertid");
-          QueryParamsMap paramsMap = req.queryMap().get("storeId");
-          String storeId = null;
-          if (paramsMap.hasValue()) {
-            storeId = paramsMap.value();
-          }
+          String storeId = req.params(":storeId");
           Metacard versionMetacard = util.getMetacardById(revertId);
 
           List<Result> queryResponse = getMetacardHistory(id, storeId);
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
@@ -372,12 +372,12 @@ public class MetacardApplication implements SparkApplication {
         "/history/:id",
         (req, res) -> {
           String id = req.params(":id");
-          QueryParamsMap paramsMap = req.queryMap().get("sourceId");
-          String sourceId = null;
+          QueryParamsMap paramsMap = req.queryMap().get("storeId");
+          String storeId = null;
           if (paramsMap.hasValue()) {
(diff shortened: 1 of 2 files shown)


ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -1146,6 +1146,21 @@ private boolean isChangeTypeDate(AttributeChange attributeChange, Metacard resul
...
 
   private List<Result> getMetacardHistory(String id) {
+    Filter resourceFilter = filterBuilder.attribute(Core.ID).is().equalTo().text(id);
+    Metacard resource = null;
+    try {
+      List<Result> results =
+          catalogFramework.query(new QueryRequestImpl(new QueryImpl(resourceFilter))).getResults();
+      if (results.isEmpty()) {
+        LOGGER.trace("Unable to find resource {} inside history endpoint.", id);
+        return new ArrayList<Result>();
```
a-asaad: ✏️ Alternatively, `Collections.emptyList()`. I guess you could also just return the `results` empty list, but I think the former is a bit clearer.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -1146,6 +1146,18 @@ private boolean isChangeTypeDate(AttributeChange attributeChange, Metacard resul
   }
 
   private List<Result> getMetacardHistory(String id) {
+    Filter resourceFilter = filterBuilder.attribute(Core.ID).is().equalTo().text(id);
+    List<Result> results =
+        ResultIterable.resultIterable(
+                catalogFramework::query, new QueryRequestImpl(new QueryImpl(resourceFilter)))
```
criscrasbell: ❓ If this is adding in support for the history of harvested data then how does this first catalog query work? Wouldn't it need to specify the source of the metacard as well in order to retrieve the metacard?

bdeining: If they are all stored locally, can't we use the local provider id vs. issuing a query to the framework?

cantstoptheunk: You're right, I need to have the frontend send the source id of the resource to this endpoint and use that

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -371,7 +372,12 @@ public void init() {
         "/history/:id",
         (req, res) -> {
           String id = req.params(":id");
-          List<Result> queryResponse = getMetacardHistory(id);
+          QueryParamsMap paramsMap = req.queryMap().get("sourceId");
```
criscrasbell: ✏️ Use a path parameter instead of a query param. The `metacardId` and `sourceId` are a candidate key for the record and should both be path params because they are both needed to retrieve the resource.

https://stackoverflow.com/questions/30967822/when-do-i-use-path-params-vs-query-params-in-a-restful-api

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/metacard-history/container.tsx
```diff
@@ -54,7 +54,12 @@ class MetacardHistory extends React.Component<Props, State> {
   loadData() {
     setTimeout(async () => {
       const res = await fetch(
-        `./internal/history/${this.model.get('metacard').get('id')}`
+        `./internal/history/${this.model
+          .get('metacard')
+          .get('id')}?sourceId=${this.model
+          .get('metacard')
+          .get('properties')
+          .get('source-id')}`
```
criscrasbell: ❗ This will not work with harvested data since the source id for harvested data is stored on a different attribute.

criscrasbell (changes requested):

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/metacard-history/container.tsx
criscrasbell: Tested this while fixing a bug in another area. The `source-id` property did contain the harvested source's id. Not sure if accessing it this way will not work however it's definitely something that needs to be tested.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
cantstoptheunk: Right, but the sourceId acts as a filter, not a specific resource identifier, meaning that the sourceId is a property of the resource

vinamartin: ❗ I would use storeIds not sourceIds. See how we're doing this for patch.

jlcsmith (approved):

criscrasbell (approved):

jlcsmith (approved):

jlcsmith:
🎉 hero success 🎉 
verified creation of history nodes

PR #367, opened 2020-09-16
[3.6.x] Update package.json, fix ts errors

jordanwilking:
These errors only show up when you `yarn start`, so I didn't catch them on the first pass.


ui-frontend/packages/catalog-ui-search/exports/multi-select-actions.tsx
```diff
@@ -12,6 +12,8 @@
  * <http://www.gnu.org/licenses/lgpl.html>.
  *
  **/
+
+// @ts-nocheck ts-migrate(2307) FIXME: Cannot find module
```
jordanwilking, line 16: 📓 The file that's being imported/exported here doesn't exist. Can we delete this file?

ui-frontend/packages/catalog-ui-search/exports/sharing.tsx
```diff
@@ -12,8 +12,9 @@
  * <http://www.gnu.org/licenses/lgpl.html>.
  *
  **/
+// @ts-nocheck ts-migrate(2307) FIXME: Cannot find module
```
jordanwilking, line 15: 📓 The file that's being imported/exported here doesn't exist. Can we delete this file?

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/result-sort/presentation.tsx
```diff
@@ -29,7 +29,6 @@ type Props = {
 const render = ({ removeSort, saveSort, hasSort, collection }: Props) => {
   return (
     <div className="min-w-120">
-      {/* @ts-expect-error ts-migrate(2322) FIXME: Property 'collection' does not exist on type 'Intr... Remove this comment to see the full error message */}
       <SortSelections collection={collection} />
```
jordanwilking, line 32: 📓 Added the ts-nocheck because putting the error above this line doesn't work. Also, this component doesn't have a `collection` property on it. Wasn't sure how we'd want to fix this.

PR #380, opened 2020-09-23
Cherry pick of fixes from old master branch

jlcsmith:
See commit history
Don't squash commits on merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,11 +3,11 @@
 Welcome to ddf-ui, the home of Intrigue. 
 
 ## Installing in DDF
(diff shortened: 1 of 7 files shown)


andrewkfiedler (approved):

andrewkfiedler:
I think there are still issues we need to fix with BBOX when it comes to USNG, it didn't look like it was updating to me, but I didn't see it failing either.  UTM was weird, it would work but would jump to DD during drawing then return to UTM after.

Overall looks like everything still works though.  ✅

PR #402, opened 2020-10-05
Update location editing via Managed Attributes to use the specialized location input

a-asaad:
A location entry is now used when editing a metacard's location through the "manage attributes" modal (accessed from the Inspector's summary tab), as opposed to a plain text input. This PR also fixes an issue where all Lat/Lon (DMS) entries were considered invalid.

Screen Shot 2020-10-05 at 10 12 27 AM

Screen Shot 2020-10-05 at 10 12 59 AM

Testing Instructions:
1. Build `DDF-2.26.0`, once the distribution is unzipped and started, run `profile:install standard` from the Karaf console.
2. Build this PR and follow the instructions in https://github.com/codice/ddf-ui/blob/master/README.md to install the catalog-ui.
3. Upload a geo-enabled resource.
4. Search for the resource, open it in the inspector, then choose "Manage Attributes" from the summary tab.
5. When you click the edit icon on the `location` field, you should get a location input, experiment with setting the location to different values.
  * Things to look out for:
    - The Lat/Lon (DMS) input should be working properly
    - The save button should be disabled when the input is invalid.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-new/utils/dms-utils.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-new/utils/dms-utils.js
@@ -102,8 +102,14 @@ function dmsCoordinateToDD(coordinate) {
 function dmsPointToWkt(point) {
   const latitude = parseDmsCoordinate(point.latitude.coordinate)
   const longitude = parseDmsCoordinate(point.longitude.coordinate)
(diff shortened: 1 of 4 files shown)


vinamartin (approved):

jordanwilking (approved):

andrewkfiedler (approved):

andrewkfiedler:
Verified that the enhanced input is available, and that updating location on the metacard works properly in DMS mode.  Nice touch disabling save when the location isn't valid.  🎉

PR #405, opened 2020-10-08
[3.4.x] G-8885; Use window.event instead of event

cassandrabailey293:
forward port of https://github.com/codice/ddf/pull/6368

_______________________________________________________________________

#### What does this PR do?
This PR fixes an issue where the anyGeo component was jumping off the search form in a specific version of Firefox. The problem was the usage of `event` in the `padEndWithZeros` method, which was being called many times by the pipe function. I removed the pipe function since we only want to call `padEndWIthZeros` when switching to the DMS tab from other tabs.

#### Who is reviewing it? 

@hayleynorton @andrewzimmer @zta6 @abel-connexta 

@codice/ui 

0. build, profile/install
1. Verify no regression with DMS coordinate entry/switching tabs
2. Download Firefox 60 ESR: https://ftp.mozilla.org/pub/firefox/releases/60.0esr/mac/en-US/
3. Go to the UI and Advanced Search -> anyGeo
4. Select Bounding Box
5. Switch to the DMS tab
6. Verify that there is no `event is not defined` error in the console and that the component works as expected
7. Repeat for Point Radius

Fixes: #

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

#### Notes on Review Process

#### Review Comment Legend:

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-new/inputs/masked-text-field.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-new/inputs/masked-text-field.js
@@ -26,13 +26,16 @@ class MaskedTextField extends React.Component {
     // Sometimes the event variable is defined, other times it's undefined.
     // We must capture the event in a variable when it's defined
     // in order to leverage it in checks below even when it's undefined.
-    if (event) {
-      this.prevEvent = event
+    if (window.event) {
+      this.prevEvent = window.event
+    } else if (this.prevEvent === undefined) {
+      return value
     }
     if (
       value === undefined ||
       !value.includes('.') ||
-      (((event && event.type === 'input') || this.prevEvent.type === 'input') &&
+      (((window.event && window.event.type === 'input') ||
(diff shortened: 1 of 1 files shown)


mojogitoverhere (approved):

hayleynorton:
So this will leave the underscore placeholders at all times unless the coordinate system is switched then switched back? Including when the search is executed?

andrewzimmer (approved):

cassandrabailey293:
> So this will leave the underscore placeholders at all times unless the coordinate system is switched then switched back? Including when the search is executed?

No, because in all those other situations the onBlur event will be fired. Since we're still building the DMS string each time the onBlur is called, and we're also fixing it when we're switching to the DMS tab. If I understand correctly, those are the two situations in which we need to "pad" / replace the placeholder "_" characters.

abel-connexta (approved):

zta6 (approved):

Lambeaux (approved):

shaundmorris:
build now

pklinef (approved):

PR #407, opened 2020-10-10
Add browser incompatible page for IE

rfding:
Displays a browser incompatible page when users try to access the system from an incompatible browser.

Supported browsers:
- Microsoft Edge, versions 80+
- Mozilla Firefox, versions 72+
- Google Chrome, versions 80+

(Supported versions were chosen by finding the version released closest to the start of 2020)

Unsupported browsers:
- Internet Explorer
- by default, anything else

image
Regular size browser

image
Small browser

Testing Instructions:
1. Build and run DDF.
1. Build this PR and follow the instructions in https://github.com/codice/ddf-ui/blob/master/README.md to install the catalog-ui.
1. Open in Internet Explorer.
1. Verify that the above page displays instead of the usual UI.
1. Repeat with all unsupported versions.
1. Verify that the usual UI displays in supported versions.


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/incompatible-browser.less
```diff
@@ -0,0 +1,53 @@
+@import '../../styles/vars.less';
```
rfding: I put this import here so I could use predefined settings even though I'm not running `ApplicationSetup.js`. Please let me know if there's a better way to do it.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/incompatible-browser.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react'
+const NoIE = require('../../img/no-ie.svg') as string
+const Edge = require('../../img/edge.svg') as string
+const Firefox = require('../../img/firefox.svg') as string
+const Chrome = require('../../img/chrome.svg') as string
+
+export const IncompatibleBrowser = () => {
```
rfding: I wasn't sure where the best location to put these files would be. Is `component/pages` okay or is there a better place to put them?

vinamartin: 👍  I feel like this is as fine a place as any.

vinamartin (approved):
Not 100% sure about the less thing so I'll defer to the other reviewers for that, but other than that LGTM

lifenouveau (approved):
I almost wonder if we should have a click through if a user wants to proceed anyway in case there is a future browser that might be completely compatible since we have a whitelist rather than a blacklist. 🤔 

Otherwise, the page looks great! ✨

rfding:
> I almost wonder if we should have a click through if a user wants to proceed anyway in case there is a future browser that might be completely compatible since we have a whitelist rather than a blacklist. 🤔
> 
> Otherwise, the page looks great! ✨

I can definitely make it a blacklist instead of a whitelist. That's another thing that I was unsure about.

ui-frontend/packages/catalog-ui-search/src/main/webapp/extension-points/extension-points.tsx
```diff
@@ -30,7 +30,7 @@ export type ExtensionPointsType = {
     user: any
     editableAttributes: string[]
   }) => boolean | undefined
-  customEditableAttributes: () => string[]
+  customEditableAttributes: () => Promise<any>
```
rfding, line 33: Changed this for https://github.com/connexta/aus/pull/1033 and double checked with @vinamartin as she had written the original code

andrewkfiedler:
Hero:
Worked appropriately on IE9, but I had issues testing on IE8 and below. 

We might want to consider inlining the incompatible html into the index.html, that way it doesn't depend as much on the javascript runtime going smoothly.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/incompatible-browser.less
andrewkfiedler: I think we should inline the styles if possible (feel free to hardcode the values rather than use the theming) since that's most likely to be compatible.

rfding: I added inlined styles for IE 9 and below, but IE 10 and 11 are still able to use this.

ui-frontend/packages/catalog-ui-search/src/main/webapp/index.html
```diff
@@ -29,19 +29,98 @@
...
+              <label style="font-size: 12px;">Google Chrome</label>
+            </li>
+          </ul>
+        </div>
+      </div>
     </div>
-    <div style="color:yellow">
-        Consider using the <a href="/search/simple">Simple Search UI</a> if IE 8 or below is your
-        only option.
+    <![endif]-->
```
rfding, line 94: This whole section above is pretty much duplicate code of the section below. 😅
The only reason I put this here is because IE 9 and below completely bypass the Javascript, so I can't rely on browser detection to change the `display` from `hidden` to `block`.

ui-frontend/packages/catalog-ui-search/src/main/webapp/index.html
```diff
@@ -29,19 +29,98 @@
...
+              <img src="./img/firefox.png" alt="firefox" style="margin-right: 24px;" />
+              <label style="font-size: 12px;">Mozilla Firefox</label>
+            </li>
+            <li class="browser">
+              <img src="./img/chrome.png" alt="chrome" style="margin-right: 24px;" />
+              <label style="font-size: 12px;">Google Chrome</label>
+            </li>
+          </ul>
+        </div>
+      </div>
```
rfding, line 105: This whole section above is pretty much duplicate code of the section below. 😅
The only reason I put this here is because IE 9 and below completely bypass the Javascript, so I can't rely on browser detection to change the `display` from `none` to `block` and I couldn't think of a better way to do it. Please let me know if you have any suggestions!

ui-frontend/packages/catalog-ui-search/src/main/webapp/index.html
```diff
@@ -29,19 +29,98 @@
...
+                padding: 20px;
+            }
+
+            #incompatible-browser li {
+                width: 176px;
+                display: inline-block;
+                vertical-align: top;
+                margin: 10px;
+                zoom: 1;
+                *display: inline;
```
andrewkfiedler, line 72: Is there supposed to be a * in front of display?

andrewkfiedler:
Rehero:
Firefox: ✅ 
Chrome: ✅ 
Latest Edge:  Got the browser incompatible page.  Tried to update, which it did, but it still showed the incompatible page.
IE11: Spinner
IE10: Spinner
IE9: Console Error
IE8: Console Error
IE7: Console Error
IE5:  ✅

ui-frontend/packages/catalog-ui-search/src/main/webapp/index.html
rfding: Yup, it's a hack to get the list to look like `inline-block` in IE 6 and 7, because `inline-block` doesn't seem to be properly supported.

andrewkfiedler:
I have an idea for skipping the bundle.js altogether that might fix some of the issues we're seeing.  Going to 😴  but I'll tagup with you in the morning or afternoon and see what you think.

andrewkfiedler:
Okay, here's what I think we should do.  Basically, scrap everything and just update the index.html to this:

```
<!DOCTYPE html>
<!--
/**
 * Copyright (c) Codice Foundation
 *
 * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
 * version 3 of the License, or any later version.
 *
 * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
 * <http://www.gnu.org/licenses/lgpl.html>.
 *
 **/
 -->

<html class="no-js">

<head>
    <meta charset="utf-8">

    <title>Intrigue</title>
    <meta http-equiv="x-ua-compatible" content="IE=Edge">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <meta name="description" content="">
    <meta name="author" content="">

    <link id="favicon" rel="icon" type="image/png" href="../../data:image/png;base64,">
    <script>
      // hacky but working way to suppress certain html (this is the only foolproof method working on all browsers, including ancient ie versions)
      // things like setAttribute, className etc just aren't guaranteed to work, so stick to using this!
      window.suppressNextTag = function() {
        document.write("<script type='application/x-suppress'>");
      };
      // feature detection as opposed to browser sniffing, lookBehind is only supported in relatively recent releases of popular browsers
      window.isLookBehindSupported = false;

      try {
          new RegExp("(?<=)");
          window.isLookBehindSupported = true;
      } catch (err) {}
    </script>
</head>

<body>
    <header>
        <style type=text/css>
            #incompatible-browser-bg {
                position: fixed;
                top: 0;
                left: 0;
                text-align: center;
                line-height: 30px;
                width: 100%;
                height: 100%;
                background-color: #0b1821;
                overflow: auto;
                font-family: 'Open Sans', arial, sans-serif;
            }

            #incompatible-browser {
                display: table-cell;
                vertical-align: middle;
                padding: 20px;
            }

            #incompatible-browser li {
                width: 176px;
                display: inline-block;
                vertical-align: top;
                margin: 10px;
                zoom: 1;
                *display: inline;
            }
        </style>
    </header>
    <script>
      // this will suppress the incompatible browser element if we're on a compatible browser
      if (window.isLookBehindSupported) {
        window.suppressNextTag();
      }
    </script>
    <div id="incompatible-browser-bg" style="color: white;">
      <div style="display: table; width: 100%; height: 100%;">
        <div id="incompatible-browser">
          <h1 style="font-size: 20px; margin-bottom: 20px;">Browser Unsupported</h1>
          <p style="font-size: 16px;">
            Please use one of the following supported browsers to access the
            system.
          </p>
          <ul id="supported-browsers" style="list-style-type:none; padding: 0px">
            <li class="browser">
              <img src="./img/edge.png" alt="edge" style="margin-right: 24px;" />
              <label style="font-size: 12px;">Microsoft Edge</label>
              <h1>Versions 80+</h1>
            </li>
            <li class="browser">
              <img src="./img/firefox.png" alt="firefox" style="margin-right: 24px;" />
              <label style="font-size: 12px;">Mozilla Firefox</label>
              <h1>Versions 72+</h1>
            </li>
            <li class="browser">
              <img src="./img/chrome.png" alt="chrome" style="margin-right: 24px;" />
              <label style="font-size: 12px;">Google Chrome</label>
              <h1>Versions 80+</h1>
            </li>
          </ul>
        </div>
      </div>
    </div>
    <script>
      // this will suppress the loading element if we're on an incompatible browser
      if (window.isLookBehindSupported === false) {
        window.suppressNextTag();
      }
    </script>
    <div id="loading" class="clearfix is-open">
        <div class="loader is-critical-animation"></div>
        <div class="loading-welcome">
            Loading <span class="welcome-branding is-bold"></span> <span class="welcome-branding-name"></span>
        </div>
    </div>
    <div id="router">
    </div>
    <iframe name="autocomplete" style="display:none" src="./blank.html"></iframe>
    <footer></footer>
    <script>
      // this will suppress the first webpack bundle from loading, if we're on an incompatible browser
      if (window.isLookBehindSupported === false) {
        window.suppressNextTag();
      }
    </script>
</body>

</html>
```

Overall there were two things I realized.  One, is that we should do feature detection as opposed to browser sniffing since chrome is deprecating the latter anyways.  The look behind regex is only supported in recent browsers (https://caniuse.com/?search=look%20behind), and we can update what feature we require as we move forward in the future.  

The second, is that we want to prevent the rest of the app from loading (without complicating the app code).  So I found a technique that suppresses the very next tag in html.  The idea is taken from [this stack overflow](https://stackoverflow.com/questions/4726362/use-javascript-to-prevent-a-later-script-tag-from-being-evaluated) .  It's definitely funky, but it works very well (and is compatible with every browser).

> The key to the solution is the document.write line. It will inject into the HTML <script type='application/x-suppress'> between the close script tag of first.js and the open script tag of second.js.

> The parser will see this and start a new script element. Because the type attribute has a value which is not one that the browser recognises as being JavaScript, it will not attempt to run its content. (So there are an infinite number of possible type attribute values you could use here, but you must include a type attribute, as in its absence, the browser will assume that the script's content is JavaScript.)

> The second.js script's opening tag will then be parsed as text content of the new script element and not executed. Finally the second.js script's closing tag will be re-purposed to close the new script element instead, which means that the remainder of the HTML is parsed correctly

Now, older browsers are so finicky, that instead of hiding / showing certain things like the loader or incompatible browser element (ie7 likes to ignore setAttribute), we just stick to this method to hide / show things (in addition to preventing the bundle js from loading).  That lets us simplify the code quite a bit.

I went ahead and tested this technique across all the browsers, even back to ie5, and it works.  🎉

andrewkfiedler (approved):

PR #412, opened 2020-10-13
Add optional param for getting a metacard w/ storeId

vinamartin:
To test: 
Using postman, make sure /internal/metacard/${id} is still working without an optional param.
Make sure /internal/metacard/${id}?storeId=${storeId} works.

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
@@ -259,7 +259,7 @@ public class MetacardApplication implements SparkApplication {
         "/metacard/:id",
         (req, res) -> {
           String id = req.params(":id");
-          String storeId = Optional.ofNullable(req.queryParams("storeId")).orElse(null);
+          String storeId = req.queryParams("storeId");
           return util.metacardToJson(id, storeId);
         });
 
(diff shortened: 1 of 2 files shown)


criscrasbell (commented):
Instead of intentionally using null I would use an overloaded method where possible.

vinamartin:
@bellcc Good call, I'll refactor 👍

criscrasbell (approved):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -259,7 +259,8 @@ public void init() {
         "/metacard/:id",
         (req, res) -> {
           String id = req.params(":id");
-          return util.metacardToJson(id);
+          String storeId = Optional.ofNullable(req.queryParams("storeId")).orElse(null);
```
glenhein: Why are you using an Optional? It looks like you are saying "If this value is null, then return null".

vinamartin: We used this for patch so I assumed it was a safeguard against getting an NPE if that param didn't exist, but I'm not super familiar with how optional params work.

glenhein: I don't think the use Optional is doing anything for you.

glenhein: The `Optional.ofNullable(req.queryParams("storeId"))` returns a populated Optional if `req.queryParams("storeId")` is non-null. It returns an Optional.empty() otherwise. The `orElse` extracts the value in the Optional, or returns null if the Optional is empty. You can replace the entire thing with `req.queryParams("storeId")`

glenhein (approved):

stevenmalmgren:
### **🥇HERO SUCCESSFUL🥇**

Sent a request with:
no source specified, local resource returned as expected 
source the resource belonged to, resource returned as expected
and one when the source did not exist, resource did not return

PR #413, opened 2020-10-13
Fix timezones with new datepicker

andrewkfiedler:
- The new datepicker for blueprint does not explicitly handle timezones.  It internally uses the JavaScript Date, so we need to hand it a value that takes this into account.  In addition, we need to update how we respond to changes to take this into account.  This is all encapsulated in the new date-helpers file.  It could also be used for other libraries like D3 that are timezone agnostic.
 - Brings some downstream changes to the timezone picker upstream, so that it's easier to use.


ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/time-settings/time-format-picker.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react'
+import { hot } from 'react-hot-loader'
+import Autocomplete from '@material-ui/lab/Autocomplete'
+import { TextField } from '@material-ui/core'
```
jordanwilking: ✏️ `core/TextField`

jordanwilking (approved):
🎉

vinamartin (approved):
💯

jordanwilking:
📓: I'm not sure the following should hold this back. There are bound to be quirks when we have to force timezones into timezone-less containers.

After working with this a bit there are a few minor things that I've noticed:

- The picker's time is always in 24 hour mode - even when the user's format is 12 hour mode
image
- The single date field is always empty to start (the between field start at the current time)
image
- Swapping from After/Before -> Between -> After/Before results in an invalid date because the Between's value doesn't have the same shape as the After/Before's value.
image

andrewkfiedler:
We'd have to add in the capability to do more than 24 hr mode to blueprint's timepicker if we want that unfortunately.  The other ones improvements we should tackle, although I'm ambivalent on whether we do them now or in a future PR.

jordanwilking:
I thought that time picker was more or less agnostic. If I change my "15:02" to "15:03", then my time goes from "3:02" to "15:03". 

I think it could get pretty complex to fix this though. I'm beginning to think that we're opening the door to a lot complaints by allowing timezone selection, but it is what it is.

jordanwilking:
As for the other issues, none of them are breaking, so maybe we should just push them off to another ticket? I'll see how much longer it takes to swap out the other pickers and move these to a new ticket if I don't get to them.

PR #415, opened 2020-10-14
Add timezone-compatible date pickers

jordanwilking:
Add the timezone-compatible pickers and utilities that Andrew made as well as one with a MUI TextField. Replaces KeyBoardDatePickers.

The first commit has already been reviewed in #413 

Instructions for testing are downstream

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-time-picker.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-time-picker.tsx
@@ -1,0 +1,176 @@
+import * as React from 'react'
+import { useState, useRef } from 'react'
+import {
+  DatePicker,
+  IDateInputProps,
+  IDatePickerProps,
+} from '@blueprintjs/datetime'
+import { DateHelpers } from './date-helpers'
+import { MuiOutlinedInputBorderClasses } from '../theme/theme'
+import TextField, { TextFieldProps } from '@material-ui/core/TextField'
+import InputAdornment from '@material-ui/core/InputAdornment'
+import IconButton from '@material-ui/core/IconButton'
+import CalendarIcon from '@material-ui/icons/Event'
+import Popover from '@material-ui/core/Popover'
+import useTimePrefs from './useTimePrefs'
+
+type DateFieldProps = {
+  value: string
+  onChange: (value: string) => void
+  disabled?: TextFieldProps['disabled']
+  variant?: TextFieldProps['variant']
+  error?: TextFieldProps['error']
+  helperText?: TextFieldProps['helperText']
+  label?: TextFieldProps['label']
+  disableShortcuts?: boolean
+  margin?: TextFieldProps['margin']
+  style?: React.CSSProperties
+  /**
+   * Override if you absolutely must
+   */
+  BPDateProps?: IDatePickerProps
+}
+
+const validateShape = ({ value, onChange }: DateFieldProps) => {
+  if (DateHelpers.Blueprint.commonProps.parseDate(value) === null) {
+    onChange(new Date().toISOString())
+  }
+}
+
+const formatInputValue = (value: string) => {
+  if (value) {
+    const date = DateHelpers.Blueprint.commonProps.parseDate(value)
+    return DateHelpers.Blueprint.commonProps.formatDate(date!)
+  } else {
+    return '' // This allows us to display an empty state without label overlap
+  }
+}
+
+/**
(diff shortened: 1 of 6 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-time-picker.tsx
```diff
@@ -0,0 +1,176 @@
...
+  React.useEffect(() => {
+    if (isInternalChange) {
+      setIsInternalChange(false)
+    } else {
+      // If the value is updated externally (i.e. a new value is passed into the component),
+      // then we need to update the display value.
+      // This allows us to keep the displayValue and value separate
+      // and type in the TextField without being hijacked
+      setDisplayValue(formatInputValue(value))
+    }
```
jordanwilking: 📓 Before this, I had to deal with either:

1. Updating the calendar view wouldn't update the text field, or
1. The text field value would update as I typed

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/useTimePrefs.tsx
```diff
@@ -0,0 +1,21 @@
+import { useState, useEffect } from 'react'
+import { useBackbone } from '../selection-checkbox/useBackbone.hook'
+// @ts-ignore ts-migrate(7016) FIXME: Could not find a declaration file for module '../s... Remove this comment to see the full error message
+import user from '../singletons/user-instance'
+
+const useTimePrefs = () => {
+  const { listenTo } = useBackbone()
+  const [, setForceRender] = useState(Math.random())
```
andrewkfiedler, line 8: Had no idea you could do this (leave out the first param).  🤯

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date.tsx
```diff
@@ -37,20 +34,10 @@ const validateShape = ({ value, onChange }: DateFieldProps) => {
 }
 
 export const DateField = ({ value, onChange, BPDateProps }: DateFieldProps) => {
-  const { listenTo } = useBackbone()
-  const [forceRender, setForceRender] = React.useState(Math.random())
+  useTimePrefs()
```
andrewkfiedler, line 40: Brilliant.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/useTimePrefs.tsx
vinamartin: ☝️ Same here lol, nice stuff

jordanwilking: I've had to learn to keep Typescript happy 😂

vinamartin (approved):
10/10 👏

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/useTimePrefs.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  useEffect(() => {
+    listenTo(
+      user.getPreferences(),
+      'change:dateTimeFormat change:timeZone',
+      () => {
+        setForceRender(Math.random())
+      }
+    )
+  }, [])
```
stevenmalmgren, line 18: Do you need to stop listening to the model here?

stevenmalmgren (changes requested):
See comment

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/useTimePrefs.tsx
jordanwilking: @andrewkfiedler, I noticed that we don't stop listening in many places when using listenTo. Should we be doing that?

useBackbone does have a cleanup callback that stops listening, but I don't understand 100% what we're doing there. It looks like we create a new backbone model and alter its listenTo to have our backbone model as its `this1, but I'm not sure how that plays into the cleanup.

andrewkfiedler: We don't need to, the stopListening is available if necessary, but the useBackbone effect cleans up the most common usecase which is unlistening on destruction of the component.

andrewkfiedler:
Hero Successful, tested both the advanced search input and the new input seen in the summary view (tried it on those without a date and those with a date).  User preferences are taken into account on both 🎉

PR #422, opened 2020-10-21
Converted user profile button data-id to static

georgeshehata:
George Shehata from ISRI team


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/app/app.tsx
```diff
@@ -529,7 +529,7 @@ const App = ({
                             className={`group-hover:opacity-100 opacity-25 relative py-3 hover:opacity-100 focus-visible:opacity-100 transition-opacity`}
                             Icon={PersonIcon}
                             iconPosition="start"
-                            expandedText={userInstance.getUserName()}
+                            expandedText="user-profile"
```
andrewkfiedler, line 532: Do we mean to have this change?  This will make the expanded menu show 'user-profile' for each user rather than their username.

andrewkfiedler (changes requested):
Not sure we want the expandedText change or not.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/app/app.tsx
dx-leidos: @georgeshehata seems like a misplacement

georgeshehata:
it seems that adding a optional parameter will be a more appropriate solution.

PR #426, opened 2020-10-23
Add timezone support for timeline

jordanwilking:
- Migrates the timeline from atlas
- The timeline labels and x-axis now reflect the user's timezone.
- Changes to the user's format and timezone cause the timeline to update
- Migrates other code from atlas and removes the dependency

Note: there's a quirk with format comparisons that results in special x-axis labels to fall on times other than midnight. 

To test:
- Ingest resources
- Navigate to search and open the timeline and inspector
- Search so your items show up on the timeline
- Use the item's creation time to sanity check the timeline's values
- Check the hover label, brush labels (click and drag), and confirm that changing the user's timezone and format settings updates the timeline (note: changing the format should only change the 12 vs 24-hour display)
- Verify that dropdowns and dialogs still work as expected since that code has been migrated from atlas

image
image


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/timeline/timeline.tsx
```diff
@@ -0,0 +1,1034 @@
...
+  const heightOffset = props.heightOffset ? props.heightOffset : 0
+
+  const possibleDateAttributes = getPossibleDateAttributes(props.data || [])
+
+  const timescale = getTimescaleFromWidth(width, min, max)
+
+  const [xScale, setXScale] = useState(() => timescale)
+
+  const [xAxis, setXAxis] = useState(() =>
+    d3.axisBottom(xScale).tickSize(AXIS_HEIGHT).tickFormat(multiFormat)
```
jordanwilking, line 336: 📓 Most of this code is unchanged. I updated things to use Moment and add the tickFormat.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/timeline/util.tsx
```diff
@@ -0,0 +1,96 @@
...
+      return formatDay(date) // Wed 08
+    } else {
+      return formatWeek(date) // Feb 08
+    }
+  } else if (d3.utcYear(date) < date) {
+    return formatMonth(date) // February
+  } else {
+    return formatYear(date) // 2003
+  }
+}
```
jordanwilking, line 96: 📓 Other than updating thing to use moment, this is all that I updated in this file. This formats the x-axis.

andrewkfiedler:
Hero successful ✅ 
Verified the timeline response to timezone by having it open and swapping between timezones.  The visual updated immediately, and was correct in each instance.

PR #427, opened 2020-10-23
I-6095 / I-6060 / I-5881 - Preserves security access attributes on overwrite

vinamartin:
Previously, security access attributes were blown away on overwrite which caused users to lose access to their data. 
This preserves them on overwrite.

To test: set these security attributes on a metacard and verify they remain after overwriting using the overwrite tab


jlcsmith (approved):

austinsteffes-aus (approved):

jordanwilking:
✍️ Hero Successful ✍️ 

#### Steps

1. Ingested a resource (image) and shared it with user1 (view) and user2 (edit)
1. Overwrote the metacard with a new image
1. Confirmed that the users still had the proper access to the metacard
1. Switched back to the previous version and all of the users we able to view it
1. Switched back again. Still good.

#### Unrelated Encounters
  - The restored metacard updates its content immediately, but not its title.

PR #437, opened 2020-10-27
Fixes UI crashing issues with query basic time

vinamartin:
To test:
1. Make sure UI doesn't crash when selecting relative time query in Basic Search
2. Create a new basic query and select time, delete all of the attributes (e.g. created, modified, etc.) verify the UI doesn't crash
3. Verify that modifying time filter actually performs the query.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/home.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/home.tsx
@@ -174,6 +174,7 @@ const LeftTop = ({ selectionInterface }: { selectionInterface: any }) => {
           color="primary"
           size="small"
           onClick={() => {
+            selectionInterface.getCurrentQuery().trigger('update')
             selectionInterface.getCurrentQuery().startSearchFromFirstPage()
           }}
         >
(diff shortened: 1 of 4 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/home.tsx
```diff
@@ -174,6 +174,7 @@ const LeftTop = ({ selectionInterface }: { selectionInterface: any }) => {
           color="primary"
           size="small"
           onClick={() => {
+            selectionInterface.getCurrentQuery().trigger('update')
```
andrewkfiedler, line 177: This surprises me a bit, I think maybe we need to update query-basic similarly to query-advanced and have it directly modify / use the filter tree as the single source of truth.

vinamartin: Yeah, I agree- I'm hoping to update it 👍

andrewkfiedler:
I think we can close this in favor of https://github.com/codice/ddf-ui/pull/441?  Or it's possible we'll want to merge that and see if some of these issues still exist.  It moves query basic to depend on filter tree, so I had to change a number of things.

PR #441, opened 2020-10-29
Fixes issues with filter tree, moves query basic to use filterTree directly

andrewkfiedler:
See commits for breakdown.  Big part of this was enforcing classes, which involved making sure typescript didn't allow using the spread operator (which results in a plain object) without feeding that into a new instance of the class.

Also optimized some bits for performance since depending on the filter tree directly caused more frequent updates to it.  Notably there is a change in the home component, and we now memo the query settings.

Testing:
Go wild testing everything on the basic page, then test going to the advanced and ensuring it translates appropriately.

In advanced, test out grouping, negation, etc.


vinamartin:
🎉  Hero successful- extra kudos on fixing a lot of the crashing issues I've been seeing with query basic time.
One minor issue I saw that I also saw on my PR and have been trying to fix is that you can't "Clear All" these time properties using the "X" button **and** you also can't remove the last one. It's kind of minor but it's a pain if you only want one of the properties and have to manually click on all of the others to clear them instead of just clearing and typing what you want.
Screen Shot 2020-10-29 at 07 17 38

I think what's happening here is that we're trying to figure out what kind of filter it is on update and it doesn't work with having undefined property values. That's what I was trying to fix here: https://github.com/codice/ddf-ui/pull/437. 
I think I'm going to close out that PR in favor of writing up a patch against this one.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/keyword.spec.js
```diff
@@ -33,51 +33,51 @@ const { shallow } = Enzyme
...
-    const fetch = async () => {
-      throw new Error('unavailable')
-    }
-    const wrapper = shallow(<Keyword fetch={fetch} onError={onError} />)
-    wrapper.find('AutoComplete').prop('onChange')({
-      id: '0',
-      name: 'test',
-    })
-  })
+  // it('should fetch features on select', (done) => {
```
jordanwilking, line 41: ✏️ Should we just delete all of these tests if we don't intend to replace them now?

jordanwilking (approved):

vinamartin (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/keyword.spec.js
andrewkfiedler: I'm not sure to be honest, I left them commented out as a guide for what we should try and fix.  They've been not running for so long that I'm not entirely sure how relevant they are, but they probably have good ideas of what we should be testing in the components (the components have just changed too much). 

I originally had wanted to add some tests to ensure the filter builder didn't regress, but ran into issues just trying to render it.  Didn't have time to chase them down, but thought I'd include the change to make it possible to turn tests back on.

jordanwilking: We could always check an earlier commit to reference these, but it's not a big deal me 🤷. I'm sure we'll revisit these as we increase our testing efforts soon.

andrewkfiedler:
@vinamartin nice catch, I think I can do something similar to what I did for types (I had to make a fake '*' 😭 )

andrewkfiedler:
@vinamartin could you rehero this?  I pushed up a fix for the issue you saw, and added the ability to press enter to run a search (if within a string field on advanced search).

vinamartin:
Manually verified ✨ 
Looks good to me, merging

PR #448, opened 2020-11-04
Adds more granularity to RELATIVE temporal search and adds AROUND temporal search type

vinamartin:
### 1. [**Reorders the time input on Query basic to be more human readable**](https://github.com/codice/ddf-ui/pull/448/commits/db6c0a50000e7fdf37614514c6cdb3c10e6f7b3f)
Changes the fields to be in "attribute" "property" "value" order to read like a sentence, e.g. "Created before 02 Nov 2020"
Screen Shot 2020-11-04 at 10 52 24

### 2. [**Adds weeks to relative query**](https://github.com/codice/ddf-ui/pull/448/commits/dfbe951c89fa2db6ec705293781adeb1b9a6564b)
Weeks wasn't previously supported as an option for relative queries because it's not supported when converting to cql. This adds weeks to the dropdown and then converts it to days when it gets converted to cql. 
Screen Shot 2020-11-04 at 07 01 01

### 3. [**Adds seconds to relative query**](https://github.com/codice/ddf-ui/pull/448/commits/6c3780894e10dd3ceaf302059bed04619d83cc15)
Seconds was already supported so this just adds it in as a dropdown option and also brings the amount and unit on to the same line visually so it takes up less space.
Screen Shot 2020-11-04 at 10 54 08

### 4. [**Adds "Around" temporal filter type**](https://github.com/codice/ddf-ui/pull/448/commits/061e8e1a5910dd085c02fd276cfca6aeb9ae7225)
Adds in a new filter type, "AROUND" which gets converted to cql as a "BETWEEN" (duration).
The idea is that "AROUND" is similar to "relative" queries, except it takes a specific date + buffer. A great comparison would be doing a point radius spatial query- this is a similar idea except temporal.
Screen Shot 2020-11-04 at 07 01 27

### 5. **Renames "Relative" to "Within the Last"**
Hopefully this name update conveys to users that relative queries imply "now". Whenever this query is run, it will always be within the last "2 weeks".
Screen Shot 2020-11-04 at 07 01 54

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/filter-builder/filter.structure.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/filter-builder/filter.structure.tsx
@@ -48,7 +48,7 @@ export const serialize = {
       return ''
     }
     //Weeks is not a valid unit, so convert this to days
-    if(unit === 'w') {
+    if (unit === 'w') {
       let convertedUnit = 'd'
       let convertedLast = (parseInt(last) * 7).toString()
       return `RELATIVE(${'P' + convertedLast + convertedUnit.toUpperCase()})`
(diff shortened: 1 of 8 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/filter-builder/filter.structure.tsx
```diff
@@ -46,9 +47,27 @@ export const serialize = {
     if (unit === undefined || !parseFloat(last)) {
       return ''
     }
-    const prefix = unit === 'm' || unit === 'h' ? 'PT' : 'P'
+    //Weeks is not a valid unit, so convert this to days
+    if (unit === 'w') {
+      let convertedUnit = 'd'
+      let convertedLast = (parseInt(last) * 7).toString()
```
vinamartin, line 53: ✏️  I feel like this is safe enough, but still-might want to check if there's a moment function that does this for us.

andrewkfiedler (approved):

andrewkfiedler:
Verified both features on basic and on advanced.
Works as expected:
Screen Shot 2020-11-04 at 7 46 09 PM
Screen Shot 2020-11-04 at 7 44 59 PM

The boundaries are non inclusive as far as I could tell (< vs <=), but I'm not sure how much control we have over that (and it's always worked like that).  I think this is a great feature, excited to see it 🎉

PR #449, opened 2020-11-05
Updates package json postbuild command to create a tgz of the compiled src, adds this as a maven artifact

andrewkfiedler:
- The postbuild script in the package json of catalog-ui-search will now use npm pack to create an installable for downstream projects (no more needing to publish to npm).  We change the name slightly and move it down into dist so it doesn't affect the formatter, and we have a stable name.
 - The artifact attach section of mvn is now aware of this file, which is designated as catalog-ui-search-src just so we can differentiate it better from the jar (which is the compiled application).  The src is merely the compiled (from typescript) src.
 - Also updated to the latest ace so the dist folder will be excluded from formatting checks.


andrewkfiedler:
Rerunning CI: https://support.circleci.com/hc/en-us/articles/115014359648-Exit-code-137-Out-of-memory
Logs: https://circleci.com/api/v1.1/project/github/codice/ddf-ui/1038/output/104/0?file=true&allocation-id=5fa35ae511276633462e0067-0-build%2F42230823

andrewkfiedler:
Screen Shot 2020-11-04 at 7 17 25 PM
Verified this works on Circle CI as well

andrewkfiedler:
Once we do a release, folks in a downstream project can then update their package json to look like this:
```
    "catalog-ui-search": "http://path-to-version-of-deployed-artifact/catalog-ui-search-src.tgz",
```
Note, it might be http or https depending on where the artifact is deployed.

vinamartin (approved):

ui-frontend/packages/catalog-ui-search/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "catalog-ui-search",
   "author": "Codice",
   "description": "A frontend UI for DDF Catalog",
-  "version": "4.0.7-SNAPSHOT",
+  "version": "5.0.0",
```
jlcsmith, line 5: should this be kept in sync with project version?

jlcsmith (approved):

PR #450, opened 2020-11-05
Updates search date/date range pickers to go back as far as 1900

vinamartin:
Updates the date/date range pickers on the search page to go back as far as 1900.
To test: For each temporal type (Before, After, During, Around), try to set the query to a date before the year 2000. Make sure it gets updated properly on the UI and that the right query gets sent across to the backend (if you disable websockets, you can use the network tab to watch the cql requests).

Screen Shot 2020-11-05 at 16 48 10
Screen Shot 2020-11-05 at 16 47 33

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-helpers.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-helpers.tsx
@@ -10,7 +10,7 @@ import {
 } from '@blueprintjs/datetime/lib/esm/common/dateUtils'
 import { getDefaultMaxDate } from '@blueprintjs/datetime/lib/esm/datePickerCore'
 
-export const DefaultMinDate = new Date('Jan 1 1900')
+export const DefaultMinDate = new Date('Jan 1, 1900')
 
 export const DefaultMaxDate = getDefaultMaxDate()
 
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-helpers.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-helpers.tsx
@@ -55,14 +55,13 @@ export const DateHelpers = {
           return ''
         }
       },
-      isValid: (date: Date, minDate?: Date, maxDate?: Date) => {
+      isValid: (
+        date: Date,
+        minDate: Date = DefaultMinDate,
+        maxDate: Date = DefaultMaxDate
+      ) => {
         return (
-          date &&
-          isDateValid(date) &&
-          isDayInRange(date, [
-            minDate || DefaultMinDate,
-            maxDate || DefaultMaxDate,
-          ])
+          date && isDateValid(date) && isDayInRange(date, [minDate, maxDate])
         )
       },
     },
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-helpers.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-helpers.tsx
@@ -77,8 +77,12 @@ export const DateHelpers = {
           }
         }) as IDateInputProps['onChange']
       },
-      generateValue: (value: string) =>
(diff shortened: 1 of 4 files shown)


jordanwilking (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-range.tsx
```diff
@@ -53,6 +53,7 @@ export const DateRangeField = ({
   return (
     <DateRangeInput
       allowSingleDayRange
+      minDate={new Date('Jan 1, 1900')}
```
andrewkfiedler: Feels like these would be worth factoring out as exported variables in date helpers.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-helpers.tsx
```diff
@@ -122,7 +142,11 @@ export const DateHelpers = {
 
           if (
             momentWithOffset.isValid() &&
-            DateHelpers.Blueprint.commonProps.isValid(momentWithOffset.toDate())
+            DateHelpers.Blueprint.commonProps.isValid(
```
andrewkfiedler, line 145: If we move the min / max to be exported variables within this file, I think we should update the isValid method in this file to use those as the defaults.

andrewkfiedler (changes requested):
See comments.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-range.tsx
vinamartin: Agreed, thought that like a few minutes after I put the PR up.

andrewkfiedler (approved):

andrewkfiedler:
Seeing a strange issue where after changing my timezone to something like -7 then back to utc the datepicker swaps to this:
Screen Shot 2020-11-06 at 1 21 07 PM

andrewkfiedler:
Screen Shot 2020-11-06 at 1 34 41 PM
Screen Shot 2020-11-06 at 1 34 36 PM

Did a little investigation.  It appears that this issue crops up at exactly January 1st, 1912.  

This seems pretty minor, but probably worth talking about a bit.  It does make selecting that date seemingly impossible.

andrewkfiedler:
I know actually believe that the screenshot above is doing the right thing (I was using Africa/Abidjan which I thought was the same as UTC, see screenshot).
Screen Shot 2020-11-06 at 1 50 10 PM
However, switching to plain UTC instead yielded this:
Screen Shot 2020-11-06 at 1 49 23 PM

Upon searching up [Africa/Abidjan 1911 on google](https://www.google.com/search?client=firefox-b-1-d&sxsrf=ALeKk02BvdctKdWNxQfcJxhzVzCS-B1eWw%3A1604695875202&ei=Q7elX__yC7T29AOfpLnIDQ&q=Africa%2FAbidjan+timezone+1911&oq=Africa%2FAbidjan+timezone+1911&gs_lcp=CgZwc3ktYWIQAzoECAAQRzoKCAAQyQMQFBCHAjoCCAA6BAgAEB46BggAEBYQHjoHCAAQyQMQCjoECAAQCjoJCAAQyQMQFhAeOggIIRCgARCLAzoFCCEQoAFQpwhYqxxg6x1oAHACeACAAZsBiAHRDJIBBDEuMTOYAQCgAQGqAQdnd3Mtd2l6yAEIuAECwAEB&sclient=psy-ab&ved=0ahUKEwi_oPT85e7sAhU0O30KHR9SDtkQ4dUDCAw&uact=5) I found this page which explains why this date causes this to happen.
https://timezonedb.com/time-zones/Africa/Abidjan

" After Sunday, 31 December, 1911 11:59:59 PM:

Clocks were moved forward to become Monday, 01 January, 1912 12:16:08 AM "

So everything looks like it works correctly, even on these sort of special cases!

jordanwilking:
@andrewkfiedler I was wondering about that. There are so many timezone edge cases that they all look like bugs. One of the -4:30 timezones is like -4:00 before 2000. That one threw me off, but was also correct.

PR #454, opened 2020-11-10
Add OWASP scan step to CI builds

SmithJosh:
### Description
Adds an OWASP scan step to the CircleCI build. It stores the report so it is available as a build artifact in the CircleCI build page. The step adds about 3 minutes to CI builds.

### Reviewers
@codice/build
@codice/continuous-integration
@codice/security

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,9 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <compiler.fork>false</compiler.fork>
+    <dependency-check-maven.version>6.0.2</dependency-check-maven.version>
     <maven.deploy.plugin.version>2.8.2</maven.deploy.plugin.version>
+    <mariadb.version>2.4.1</mariadb.version>
     <javalin.version>2.5.0</javalin.version>
     <javalin.thirdparty.bundle.version>2.5.0_1</javalin.thirdparty.bundle.version>
     <kotlin-osgi.version>1.2.61</kotlin-osgi.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -281,6 +283,50 @@
...
+            <dependency>
+              <groupId>org.mariadb.jdbc</groupId>
+              <artifactId>mariadb-java-client</artifactId>
+              <version>${mariadb.version}</version>
+            </dependency>
+          </dependencies>
+          <configuration>
+            <!-- The following properties enable using a centralized nvd server -->
+            <autoUpdate>${owasp.autoUpdate}</autoUpdate>
+            <databaseDriverName>${owasp.database.driverName}</databaseDriverName>
```
shaundmorris: ❔ Does the settings file have these owasp properties defined?

shaundmorris (approved):

pom.xml
SmithJosh: Discussed this out-of-band. Conclusion: we don't need these properties because we're not running a local cache of the NVD database and the dependency-check plugin will use the public NVD database by default. The downsides are:
1. Performance: using the local DB would save some time. On my machine (with a cache), it takes ~1 minute, whereas CircleCI takes ~3 minutes for the scan
2. Availability: Occasionally the public NVD database is unresponsive and could cause the scan to fail. Using a local copy resolves that issue.

For now, I've removed the properties with the expectation that we'll use the public database. If either of the above concerns becomes an issue, we can revisit this then

andrewkfiedler:
Will this fail builds, or just report issues?

SmithJosh:
@andrewkfiedler It just reports issues

jordanwilking (approved):

SmithJosh:
Closing this as OBE. We're switching to Jenkins for CI https://github.com/codice/ddf-ui/pull/479

PR #457, opened 2020-11-17
Initialize filterTree with valid default

SmithJosh:
We expect the filterTree to have a `filters` property (see https://github.com/codice/ddf-ui/blob/master/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js#L278), but don't initialize it with one. This fixes that.

@jlcsmith @andrewkfiedler


jlcsmith (approved):

andrewkfiedler (approved):
Can we also add a protection to that line 283 where if filterTree is undefined that if statement also triggers?

SmithJosh:
@andrewkfiedler Yep added a check

andrewkfiedler:
Hero successful: ✅

PR #458, opened 2020-11-18
Performance updates in results visual and elsewhere in the app

andrewkfiedler:
See the commits for a breakdown of all the speed increases.  
Note this also has a change to use manage attributes in the results visual, and moves export to be by paging.  The final commit explains this in more detail.

I used the react perf / browser perf tools to accomplish the majority of this, using them to track down bottlenecks in various parts of the components.  If you want insight into any part of it, feel free to message me.

There's also a fix in Query.js included to make sure isOutdated updates correctly, without which paging will not work properly.


andrewkfiedler:
@vinamartin question around highlighting, is it possible to get back the entire highlight and not just a chunk?

andrewkfiedler:
Please review even though I'm merging this, just need to do a release in case we want to pull this in to a downstream project.  I can make suggested updates afterwards.

PR #459, opened 2020-11-19
Update default checking for date, and default for date-range

andrewkfiedler:
- The date validity check would never get triggered before, now it does a simple attempt at constructing a date and sees if it's invalid.
 - The date range default used before is seen as invalid on the backend since start === end.  This just subtracts a day from start to fix the issue.


vinamartin (approved):

vinamartin:
Pulling this down to test.

vinamartin:
LGTM 👍

PR #461, opened 2020-11-20
Performance improvements to transfer list, plus split list / table manage attribute config apart, plus fix for results updating in components when edited

andrewkfiedler:
- Lots of performance enhancements to transfer list
 - Splits list / table views back into having separate configs for attributes
 - Various spacing fixes for result item
 - Editing a metacard will now properly update components that register for backbone "syncs" (main one is the results visual
 - Fix sticky checkbox issue in table
 - Fix bottom most row missing border (only noticeable on tables where scrolling didn't happen, aka not many results.)


andrewkfiedler:
The performance enhancements on this I'll do try to do a short presentation / explanation on.  Some are more straight forward than others.

PR #464, opened 2020-12-01
Limits limitToDeleted search option to local resources

vinamartin:
Limits the _limitToDeleted_ search option to only return :
- resource metacards
- local/harvested results

This is ideal because we wouldn't expect archived data from federated sources (we have no means of interacting with them) and because some sources can't handle this query properly. 

**To test:**
- Configure a federated source
- Disable websockets
- Archive a variety of items (including non-resource items where possible)
- Perform a * search using the `limitedToDeleted` option (message me for instructions on how to do this)
- See that archived items come back in the query results and that the query only hit the local sources (you can see this by using the network tab and looking at the request)
- Modify your search criteria and perform another search using the option, verify it retains your criteria while also applying the limitedToDeleted option

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
@@ -65,6 +65,11 @@ function limitToDeleted(cqlFilterTree) {
         type: 'ILIKE',
         value: 'deleted',
       },
+      {
+        property: '"metacard.deleted.tags"',
+        type: 'ILIKE',
+        value: 'resource',
+      },
     ],
   }
 }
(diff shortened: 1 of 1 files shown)


jordanwilking (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
```diff
@@ -335,18 +340,25 @@ Query.Model = Backbone.AssociatedModel.extend({
     data.batchId = Common.generateUUID()
 
     // Data.sources is set in `buildSearchData` based on which sources you have selected.
-    const sources = data.sources
+    let selectedSources = data.sources
+    const harvestedSources = Sources.getHarvested()
+
+    const isHarvested = (id) => harvestedSources.includes(id)
+    const isFederated = (id) => !harvestedSources.includes(id)
```
glenhein, line 347: Could `id` ever be the local catalog, in which case it isn't harvested or federated?

glenhein (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
vinamartin: "Harvested" in this scenario always includes the local catalog. The language is a bit confusing but if you check out the getHarvested() function it will always include the local catalog.

derekwilhelm (approved):

jordanwilking:
Performed the listed steps downstream and everything worked as expected. More details can be found there.

PR #466, opened 2020-12-02
Update paging to be fuzzy

jordanwilking:
Updates the query feed to use fuzzy results

Testing:

1. Ingest some data
1. Search and note that the hits are fuzzy as well as the counts when you click the heartbeat icon
1. Verify that toggling the config disables the fuzzy results
image

image
image


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/fuzzy-results.tsx
```diff
@@ -0,0 +1,10 @@
+export const fuzzyResultCount = (resultCount: any) => {
+  if (resultCount < 10) return '< 10 hits'
```
vinamartin: 📝 Is it worth making this just be like the number without the word 'hits' so it's more extensible for other uses? (For instance if we want to use this in other places that don't require display or something). 

I think this is fine as is, just asking. 👍

vinamartin (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/fuzzy-results.tsx
jordanwilking: Good thought. I'll make that change 👍.

stevenmalmgren (approved):

jzstern (approved):

SmithJosh:
❓ Is the available/possible stat still useful? Seems like the numbers would be the same most of the time
❓ Also, out of curiosity, what was the motivation for this change?

jordanwilking:
@SmithJosh Good question and possibly not. There's a link to the ticket on the downstream pr, which talks about it a bit. Ultimately, it comes down to obfuscation for perceived security benefits.

It may be worth bringing up in the next ui tagup.

SmithJosh (approved):
Hero successful, though I noticed this [config file](https://github.com/codice/ddf-ui/blob/master/ui-backend/intrigue-ui-app/src/main/resources/etc/org.codice.ddf.catalog.ui.config) is missing the `fuzzyResultsEnabled` property, so after a clean install, the setting shows as disabled in the Admin UI, though fuzzy results are in fact enabled. Should add the property to that file.

PR #467, opened 2020-12-02
[WIP] Adjusts spellcheck to be done presearch (inline on any text field)

andrewkfiedler:
- blank and * do not send a spellcheck suggestion, we can adjust this as much as we like.
 - does not remove the old setting, this is to see if folks like this style.  If they do, we'll add on to this and use the old setting to turn on and off spellcheck
 - Right now it's a canned suggestion that simply adds 's' to everything.
 - Users can dismiss them, or accept them.

Screen Shot 2020-12-02 at 2 12 33 PM
Screen Shot 2020-12-02 at 2 13 14 PM
Screen Shot 2020-12-02 at 2 13 08 PM
Screen Shot 2020-12-02 at 2 12 45 PM

Do not merge this, it's merely for others to decide if we want to move forward on this.  We'd need to remove some old spellcheck stuff, and adjust it to use the old setting for if spellcheck is on or not (for users to turn off / on).

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/text.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/text.tsx
@@ -14,8 +14,14 @@
  **/
 import * as React from 'react'
 
-import MuiTextField from '@material-ui/core/TextField'
+import MuiTextField, {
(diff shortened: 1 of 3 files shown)


andrewkfiedler:
Closing for now until we have a decision on this.

PR #470, opened 2020-12-09
Adds migration capability to cql / filter tree

andrewkfiedler:
- Query models will now fallback to cql in order to reconstitute a filter tree appropriate for the current application version (if filterTree is null or undefined, we can add other methods to trigger this as necessary or desired).
 - CQL has been typed out as much as possible to make it so when we do update the filter tree structure, we can keep the reconstitution (migration) capability working.
 - Old cql strings can now be reconstituted into the current filter tree structure through a number of changes in the CQL file.  Also, simplification has been updated to take advantage of boolean algebra as much as possible in the process.  If we didn't simplify, the result of reconstitution would be super nested due to how postfix notation works.

To ease testing, I've added a temporary(?) route only visible when in developer mode (running the dev server).  It shows the original on the left, and the migrated from pure cql on the right (uneditable).  
Screen Shot 2020-12-09 at 1 16 48 AM


andrewkfiedler:
Needs a few typescript fixes (forgot I was using transpile for dev)

vinamartin (approved):

jordanwilking (approved):

jordanwilking:
Local build successful on Windows after `yarn format -w` at the ui-frontend level (this has been an ongoing formatting issue on Windows).

image

PR #473, opened 2020-12-11
Search Result Highlighting Update Highlight Style (#6243)

lifenouveau:
Update highlighted results to have light blue backgrounds while removing the bold effect. 

I added a transparency to these to blend better with light and dark themes: ✅ 

I thought the 100% blue highlight was overpowering the dark text: ❌ 

Light mode:
Screen Shot 2020-12-15 at 10 05 34 AM

Dark Mode:
Screen Shot 2020-12-15 at 10 08 24 AM

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/theme/theme.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/theme/theme.tsx
@@ -86,7 +86,7 @@ const GlobalStyles = createGlobalStyle<ThemeInterface>`
         outline: none !important; 
       }
       span.highlight {
-        font-weight: bolder;
+        background-color: rgba(66, 110, 203, 0.40);
       }
       .MuiToolbar-root a,
       .MuiToolbar-root .MuiBreadcrumbs-separator {


vinamartin (approved):

jordanwilking (approved):

lifenouveau:
Build now

andrewkfiedler:
The light mode feels hard to read in my opinion, I think we need better contrast there.

lifenouveau:
Should we reduce the color intensity programmatically somehow? @andrewkfiedler

andrewkfiedler:
> Should we reduce the color intensity programmatically somehow? @andrewkfiedler

Depends on if you want to have a different background for light / dark.  If they're the same maybe we could set color of text to always be white?  Not sure if that'd look great though.

Making them programmatically different is easy, so I'm open to that route if we'd rather have them be styled completely differently.

lifenouveau:
> > Should we reduce the color intensity programmatically somehow? @andrewkfiedler
> 
> Depends on if you want to have a different background for light / dark. If they're the same maybe we could set color of text to always be white? Not sure if that'd look great though.
> 
> Making them programmatically different is easy, so I'm open to that route if we'd rather have them be styled completely differently.

I tried the white text route for the active highlighted text, so I'm open to that simple route for this. I agree that even with the reduction of the color, it's still not neutral enough to be easily legible with dark text.

lifenouveau:
I've decreased the opacity to 40% which seems to be a good balance for both color modes.

PR #474, opened 2020-12-15
Updates CustomFilterInput extension point

vinamartin:
Updates CustomFilterInput to take in **value** and **onChange** to be more usable across the UI.
Previously, you wouldn't be able to use this CustomFilterInput in the various fields of "Near" so you could only use this for single input filter types.

To test, create a CustomFilterInput in a downstream project, for example here, I've just made a simple textfield with the blue coloring on the outside. Verify you can see the custom filter input for only string property queries (title, anytext, etc) and that contains shows your custom filter input. Verify IS EMPTY doesn't show your custom filter input. Verify NEAR shows your custom filter input for both fields. 
Screen Shot 2020-12-15 at 10 54 03
Screen Shot 2020-12-15 at 10 54 13


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/near.tsx
```diff
@@ -23,6 +24,36 @@ type NearFieldProps = {
   onChange: (val: ValueTypes['proximity']) => void
 }
 
+const showCustomTextFieldOrDefault = ({
```
andrewkfiedler: Is it possible to factor this out into a common component between near and the filterInput?

vinamartin: Yeah, I think so since they seem to be more or less the same thing 👍

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/filter/filter-input/customInputOrDefault.tsx
```diff
@@ -0,0 +1,39 @@
+/**
```
jordanwilking, line 1: I asked about the copyright because I wanted to get rid of these ugly things 😢

jordanwilking (approved):

jordanwilking:
📶 Hero Successful 📶 

#### Steps

- Used custom component downstream (detailed steps can be found there)

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/filter/filter-input/customInputOrDefault.tsx
```diff
@@ -0,0 +1,52 @@
...
+}: {
+  value: string
+  onChange: (e: string) => void
+  props?: Partial<TextFieldProps>
+}) => {
+  let textValue = value
+  //Clear out value when switching between structured string inputs (e.g. NEAR)
+  if (typeof textValue !== 'string') {
+    textValue = ''
+  }
```
stevenmalmgren, line 33:
```suggestion
  const textValue = typeof value !== string ?  ' ' : value
```

stevenmalmgren (approved):

PR #476, opened 2020-12-17
fix default column order for exporting the table

glenhein:
For exporting a table, the default column order was "{}", but needs to be "[]".

This is for the "Export" link that appears at the top of the search results, not the export option from the three-dot context menus.

To test, you will need two sources, one local and one remote. I used the geoserver docker and created a WFS 1.1.0 source. See the geoserver readme if you need help setting it up.

Try exporting results with two source, one local source, and one remote source.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/table-export/table-export.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/table-export/table-export.tsx
@@ -20,14 +20,12 @@ import {
   Transformer,
 } from '../../react-component/utils/export'
 const user = require('../../component/singletons/user-instance.js')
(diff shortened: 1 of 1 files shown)


SmithJosh (approved):

vinamartin (approved):

andrewkfiedler (approved):

SmithJosh:
Tested this and it works. I was able to export results from a local or remote source or a mix of results from each from the table view.

I did notice though that if you change the selected sources or change any search criteria at all, the "Export" link goes away. It will only reappear if you run another search. Idk why the export would depend on the current input on the search form.

vinamartin:
Building locally, will merge if local build passes

jordanwilking:
image

PR #477, opened 2020-12-17
Unfuzz available results

jordanwilking:
Unfuzzes the available results. 

Testing:
- Verify that the results are unfuzzed (exact counts rather than fuzzy counts)

image
image

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/query-feed.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/query-feed.tsx
@@ -207,27 +207,30 @@ const QueryFeed = ({ selectionInterface }: Props) => {
     selectionInterface,
   })
   const statusBySource = Object.values(status)
-  let resultCount = '',
+  let resultMessage = '',
     pending = false,
     failed = false
   if (statusBySource.length === 0) {
-    resultCount = 'Has not been run'
+    resultMessage = 'Has not been run'
   } else {
     const sourcesThatHaveReturned = statusBySource.filter(
       (status) => status.hasReturned
     )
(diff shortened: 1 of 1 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/query-feed.tsx
```diff
@@ -103,7 +103,7 @@ const QueryStatusRow = ({ status, query }: { status: Status; query: any }) => {
       </Cell>
       <Cell data-id="available-label">
         <CellValue
-          value={fuzzyHits(status.count)}
+          value={`${status.count} hit${status.count === 1 ? '' : 's'}`}
```
SmithJosh, line 106: +1 for correct pluralization

SmithJosh (approved):
So was the conclusion that we don't need to hide the available results count since the user can count them via paging anyway?

vinamartin (approved):

andrewkfiedler (approved):

austinsteffes-aus (approved):

jordanwilking:
> So was the conclusion that we don't need to hide the available results count since the user can count them via paging anyway?

@SmithJosh, I wasn't in that meeting, but I imagine that's part of it. Obfuscating the available results doesn't hide much and it makes that field pretty worthless in most cases.

SmithJosh (approved):
Hero successful. Verified that the "Available" stat shows the exact hit count while "Possible" is fuzzy.

Also, as an aside, imo the aggregate hit count would be more helpful if it showed the available hits
Screen Shot 2020-12-18 at 9 40 37 AM

jordanwilking:
image

PR #478, opened 2020-12-17
Make URL attributes into hyperlinks in the summary tab

jzstern:
Makes it so that attributes in the summary tab of the inspector view that are URLs are rendered as hyperlinks

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
@@ -79,12 +79,12 @@ const ThumbnailInput = ({
           type="file"
           ref={fileRef}
           style={{ display: 'none' }}
-          onChange={e => {
+          onChange={(e) => {
             if (imgRef.current === null) {
               return
             }
             const reader = new FileReader()
-            reader.onload = function(event) {
+            reader.onload = function (event) {
               try {
                 // @ts-ignore ts-migrate(2531) FIXME: Object is possibly 'null'.
                 onChange(event.target.result)
@@ -198,7 +198,7 @@ export const Editor = ({
                         disableClearable
                         size="small"
                         options={enumForAttr}
-                        renderInput={params => (
+                        renderInput={(params) => (
                           <TextField {...params} variant="outlined" />
                         )}
                       />
@@ -209,7 +209,7 @@ export const Editor = ({
                       return (
                         <DateTimePicker
                           value={val}
-                          onChange={value => {
+                          onChange={(value) => {
                             values[index] = value
                             setValues([...values])
                           }}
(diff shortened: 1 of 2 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
```diff
@@ -432,153 +435,179 @@ const AttributeComponent = ({
...
+                                      />
+                                    </Typography>
+                                  )
+                                }
+                                return displayHighlightedAttrInFull(
+                                  lazyResult.highlights[attr],
+                                  val,
+                                  index
+                                )
+                              } else if (isUrl(val)) {
```
vinamartin: ❓ I think this is fine as is, but should we do something special if it is a link and also highlighted? I'll defer to @lifenouveau here.

vinamartin (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
lifenouveau: Theoretically the background would change, right?

vinamartin: Taking a look at this logic again, it only shows it as a url if it isn't highlighted. 🤔

lifenouveau: I see, ideally we have both.

jzstern: Ahhh I didn't realize highlighted meant exactly what you'd think it means. Will refactor so it links it for highlighted attributes as well.

jzstern: There's an issue where the <a> tag overwrites the inner highlight formatting, so I'm highlighting the entire link

lifenouveau (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
```diff
@@ -432,153 +435,191 @@ const AttributeComponent = ({
   const { isNotWritable } = useCustomReadOnlyCheck()
   const dialogContext = useDialog()
 
+  const isUrl = (value: any) => {
+    if (value && typeof value === 'string') {
+      const protocol = value.toLowerCase().split('/')[0]
+      return protocol && (protocol === 'http:' || protocol === 'https:')
+    }
```
vinamartin, line 439: If value doesn't exist or isn't a string, `return false`

vinamartin:
✅ Manually verified LGTM

PR #481, opened 2020-12-18
Fixes a race condition where extension point was referenced too early

vinamartin:
Changes the order ever so slightly of this call out to our extension point to avoid a race condition where the extension point would never get called.

To test: 
Using a downstream project, create a customEditableAttributes extension and verify that the customEditableAttributes extension point gets called.


stevenmalmgren (approved):

jlcsmith (approved):

astagg (approved):

SmithJosh (approved):
Hero successful. Verified that the customEditableAttributes extension point gets called and the editable attributes are indeed editable.

PR #482, opened 2020-12-21
Fixes ids for map layers

vinamartin:
[This function](https://github.com/codice/ddf-ui/blob/master/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/layers/layers.view.js#L130-L141) was looking for the data-id in order to properly update the order but it was failing since the data-id would never match any of the ids.

To test, make sure you can reorder map layers, reset default, and do all those fun map layer things.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/layer-item/layer-item.collection.view.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/layer-item/layer-item.collection.view.js
@@ -23,7 +23,8 @@ import LayerItem from '../../react-component/layer-item'
 const LayerItemView = Marionette.ItemView.extend({
   attributes() {
     return {
-      'data-id': this.model.id,
+      'data-id': 'layer-item-collection-container',
+      'layer-id': this.model.id,
     }
   },
   template() {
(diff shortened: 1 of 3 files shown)


jordanwilking (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/layer-item/layer-item.collection.view.js
```diff
@@ -23,7 +23,7 @@ import LayerItem from '../../react-component/layer-item'
 const LayerItemView = Marionette.ItemView.extend({
   attributes() {
     return {
-      'data-id': 'layer-item-collection-container',
+      'data-id': this.model.id,
```
vinamartin: Not sure if this should be something like `${this.model.id}-layer-item-collection-container` or not.

stevenmalmgren: Probably should. We'll be using this for test cafe so we'll want something to indicate what it is.

astagg (approved):

SmithJosh (approved):
Hero successful. Confirmed I could do all the fun map layer things: reordering, resetting defaults, and that the ordering of the layers was reflected properly on the map

PR #488, opened 2021-01-06
Adds saved searches

andrewkfiedler:
**in order for this to work, the system needs the latest ddf-jsonrpc release installed**
https://github.com/connexta/ddf-jsonrpc/releases/tag/ddf-jsonrpc-0.6

 Lots to unpack on this one.  The primary goal was to add saved searches back into Intrigue.  Might be easier to watch the video, because there are a lot changes in this one.

Video Part 1
https://www.youtube.com/watch?v=sclGWZYKnMw&feature=youtu.be
Video Part 2
https://www.youtube.com/watch?v=_WoKthxOmvc&feature=youtu.be

 - Adds ability to save searches to the Search route.  
 - Adds autosaving (only option, no manual saving is possible at the moment).
 - Adds AsyncTasks framework to abstract more complicated network activities from components.  There are related hooks components can use if they want to track them, based on their various properties.  Some upsides to this are we no longer have the tasks contained within a component, so destruction of the component or rerendering does not affect them.  It also centralizes them so various other views can hook in and update based on tasks.  
 - Adds a "are you sure you want to leave" popup for if an Async Task is in progress and the user tries to leave the webpage.
 - Reworking of the left pane on Search to accommodate the saving and saved searches being added.  The main difference is there is now a top and bottom to the left pane.  At the top are things related to modifying the search (name, type, open, move to trash).  At the bottom are collapse and the "execute search" button.  
 - The left pane now handles collapse a bit differently, with vertical text when it makes sense and icons when it doesn't.  It also changes the location of certain elements since they should be stationary rather than dependent on how long the name of the search is, notably the search actions dropdown moves from being to the right, to being on top when going from expanded to collapsed mode.
 - Since users might not immediately find the "new, open" options in search, these are also available on the left side bar as top level routes (create and open respectively).  This allows us to bring them a great deal of visibility and make sure users don't have issues finding them, while at the same time avoids polluting the search page route with trying to do that.  Instead, they can remain relegated within the menu there, and overtime users will realize they are available there as well.
 - Note that for now the create route is pretty sparse, but I imagine we'll be adding more and more types of things other than searches that users can add to the system.  For now, it auto pops out the list of possibilities, similar to the open and restore pages, since that will help guide users to the purpose of the page.
 - Adds a browse page for simply viewing all searches by most recently modified, primarily.  There are other things users could use this for, but it's primarily useful to examine searches in the system using the right pane as a preview area.  Multiple searches can be selected as once, but note that things like source selections aren't yet supported.  
 - Adds a special action for searches in the results list that allows editing them.  
 - Note that all the dropdowns use most recently modified as a sort, as this is the most likely what users would want, especially if they don't plan on filtering.
 - Updates Query / LazyResults to allow creation of queries that do not integrate ephemeral filters / sorts.  This caused me all sorts of headaches, because suddenly things would be in a weird order or not appear in dropdowns (since we use a query model in the background).  It makes me wonder if this has caused issues in other downstream projects.
 - Also adds a useMenuState hook that greatly simplifies usage of menus, popovers, etc.  
 - Updates how actions on hover get shown for result items in the list view.  Previously they were at the bottom, now they are at the top and flow vertically rather than horizontally.  This address some complaints about the prior ux, while still avoiding the issues that the original tried to solve (not covering the result's info, being readily accessible).


clockard (approved):
✅ Manually tested out the functionality outlined in the videos and everything worked as advertised

TonyMorrison:
build now

kentmorrissey:
I tried to rebase this PR onto `master`, but the rebase created **a** merge conflict~s~.

andrewkfiedler:
> I tried to rebase this PR onto `master`, but the rebase created **a** merge conflict~s~.

Next time try doing 
`git pull origin master`
instead of rebasing.
It tends to work much better.  Don't ask me to explain why though 😂 .  

With that technique, all I had to do was run that, then push again.  No conflicts.

andrewkfiedler:
@clockard , with the latest commit, cql should now update whenever a search is saved or created.

PR #489, opened 2021-01-06
Update useEffect dependency to fix stale callback

andrewkfiedler:
- We need to specify the onChange as a dependency for the hook, otherwise the reference will never update for the callback, even when the parent passes a new version of the function.  This was causing location inputs to revert the filter tree to the initial state it saw (first render) whenever the location input had a change.


jordanwilking (approved):

jordanwilking:
🏟️ Hero Successful 🏟️ 

#### Steps

1. In search, added non-anyGeo filters, added a geo filter, altered the non-anyGeo filters, and changed the geo filter's mode. Verified that all other filters retained their values.
2. Added a group with a text filter and anyGeo filter. Ran through the same steps without issue
3. Also tried with temporal filter
4. Verified downstream with same steps

vinamartin (approved):

PR #503, opened 2021-01-27
Update date fields to support millisecond, adds tests around date components, filter leaf components

andrewkfiedler:
- Easy enough to enable in the blueprint datepicker
 - The date helper part is needed because our process of handling timezone would inadvertantly drop the milliseconds.  This simply adds them back in.
 - Add tests around the date components, filter leaf components

Might open a backport for this depending on if downstream projects with more restricted versions want this functionality.


vinamartin (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-range.spec.tsx
```diff
@@ -0,0 +1,247 @@
...
+    user
+      .get('user')
+      .get('preferences')
+      .set('dateTimeFormat', Common.getDateTimeFormats()['ISO'])
+  })
+  it(`should not allow overlapping dates`, () => {
+    user
+      .get('user')
+      .get('preferences')
+      .set('dateTimeFormat', Common.getDateTimeFormats()['ISO'])
```
jordanwilking, line 87: ❓ Is this needed?

I'm making a few assumptions here. I'm guessing that each `it` is a new test and `beforeEach` should run before each `it` so this should already be set, right?

jordanwilking (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-range.spec.tsx
andrewkfiedler: I think we can remove it.  I'll give it a try.

jordanwilking (approved):

PR #512, opened 2021-02-04
[4.1.x] Add option to auto-pan to shapes on search and map load

jordanwilking:
- Pans to shapes on search (if option is on, else nothing)
- Pans to shapes on map load (if option is on and shapes present, else pans home)

Testing:

1. Navigate to Search
2. Verify that auto-pan is turned on
  
    

3. Draw shapes on the map and verify that clicking the search button pans to the shapes (try this with each shape and with different buffer sizes)
4. Open 2D and 3D maps to verify that the map auto-pans to your shapes.
5. Remove all of your shapes, set a home location, and open 2D and 3D maps to verify that panning defaults to home when there are no shapes.
6. Turn off auto-pan and verify that it's no longer functioning


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/map.openlayers.js
```diff
@@ -341,6 +341,29 @@ export default function (
         })
       }
     },
+    panToShapesExtent() {
+      var extent = Openlayers.extent.createEmpty()
```
stevenmalmgren, line 345: What's an extent?

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/map-settings/presentation.tsx
```diff
@@ -30,22 +38,45 @@ const Root = styled.div`
...
+          labelPlacement="start"
+          style={{ paddingLeft: '10px' }}
+        />
+      </FormGroup>
 
-      <ExampleCoordinates {...props} />
+      <div style={{ marginLeft: '-14px' }}>
+        <Enum
+          options={[
+            { label: 'Degrees, Minutes, Seconds', value: 'degrees' },
```
stevenmalmgren: Maybe throw these in an enum (datatype)?

stevenmalmgren (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/map.openlayers.js
jordanwilking: As I understand it, it's smallest bounding box that could contain all of the included geos. So, they're all distilled into four points that we can pan to.

stevenmalmgren:
Screen Shot 2021-02-08 at 2 10 46 PM
Verified auto-pan tries to include all shapes added to the filter on search (and does not when auto pan is disabled) on both 2d and 3d maps. 
One issue: When both 3d and 2d visualizations are open the auto pan button does not stay in sync across visualizations. If it is disabled on 2d it will also be disabled on 3d, but again the button will indicate it is enabled.

vinamartin (approved):

andrewkfiedler (approved):

stevenmalmgren:
### **🥇HERO SUCCESSFUL🥇**

PR #517, opened 2021-02-09
[4.1.x] Adds warning icon when a source status is not successful to heartbeat

stevenmalmgren:
This icon is added when a source's returned status is not successful. 
Screen Shot 2021-02-08 at 6 04 18 PM
 Testing:
Configure a WFS source that points to an invalid destination. This should force an error for that source when you query.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/query-feed.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/query-feed.tsx
@@ -210,7 +210,8 @@ const QueryFeed = ({ selectionInterface }: Props) => {
   const statusBySource = Object.values(status)
   let resultMessage = '',
     pending = false,
-    failed = false
+    failed = false,
+    warnings = false
   if (statusBySource.length === 0) {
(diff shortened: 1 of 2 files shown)


vinamartin (approved):

jlcsmith:
reminder to fwd port ot master

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/query-feed.tsx
```diff
@@ -275,15 +276,31 @@ const QueryFeed = ({ selectionInterface }: Props) => {
...
-                  data-id="heartbeat-button"
-                  onClick={handleClick}
-                  className="details-view is-button"
-                  title="Show the full status for the search."
-                  data-help="Show the full status for the search."
-                >
-                  <span className="fa fa-heartbeat" />
-                </Button>
+                <div>
+                  <div style={{ position: 'relative' }}>
```
andrewkfiedler: Try and use the tailwind utilities for these more common stylings.

In this case it would be:
```
className="relative"
```

jordanwilking (approved):

PR #518, opened 2021-02-10
[4.1.x] Combine secondary filter in a filter builder class.

stevenmalmgren:
If the object being passed into cql.write is not an instance of FilterBuilderClass it will not be processed by the cql.uncollapseNOTs logic - preventing the NOT from being applied to the final CQL. This was the case with the pairing of the user's search criteria and the secondary filter - ephemeralFilter in Query.js.


andrewkfiedler (approved):

stevenmalmgren:
This was already implemented in master so no need for a forward port.

jordanwilking (approved):

jordanwilking:
🌐  Hero Successful  🌐 

This one should be good since it's already in master. Did a quick test just to sanity check.

#### Steps

1. Ingested a few resources
2. NOT-ed the text of one resource with the secondary filter
3. Verified that the resource was removed from the search by the secondary filter
4. Tried the same with a NOT group and the resources were removed as expected.
image

PR #523, opened 2021-02-12
[4.1.x] Log when a user exports metacards

stevenmalmgren:
Adds a security log for when a user exports results or individual metacards.
Testing:
1. Add another source - e.g. CSW loopback
2. Upload resources
3. Export results from both sources
4. Verify that the action is logged as follows
Individual export (metacard interactions): "`{user} export a metacard with id {id} and source {source}`"
Multiple items exported (export at the top of the page or multiple selected items export): "`{user} exporting metacards: [{ids}] from sources: [{sources}] to format: {selected_format}`"

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/handlers/CqlTransformHandler.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/handlers/CqlTransformHandler.java
@@ -13,6 +13,8 @@
  */
 package org.codice.ddf.catalog.ui.query.handlers;
 
+import static java.util.stream.Collectors.toList;
+
 import com.google.common.collect.ImmutableMap;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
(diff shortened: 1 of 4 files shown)


jlcsmith:
@stevenmalmgren can you include an example of the logging of single and multi-result exports?

jordanwilking (approved):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/query/handlers/CqlTransformHandler.java
```diff
@@ -268,7 +274,21 @@ public Object handle(Request request, Response response) throws Exception {
...
+    List<String> metacardIds =
+        results.stream().map(result -> result.getMetacard().getId()).collect(toList());
+
+    Set<String> sourceIds =
+        results
+            .stream()
+            .map(result -> result.getMetacard().getSourceId())
+            .collect(Collectors.toSet());
+
+    securityLogger.audit(
```
jlcsmith: It looks like this will list the metacard Ids and source Ids but it won't be clear which metacard is from which source right?  Can we update the logging to make this clear?  I think it would be OK if this results in multiple audit log statements.

jlcsmith (approved):

vinamartin (approved):

vinamartin:
Manually verified!

PR #529, opened 2021-02-16
[4.1.x] Fix setting Point Radius USNG/MGRS radius

jordanwilking:
Fixes issue where the point radius mgrs radius field was uneditable

Testing:
1. Open 2d and 3d maps
2. Select Location -> Point-Radius and draw a point radius on the map
3. Swap to the USNG/MGRS tab and verify that you can edit the radius field and that the map updates
4. Try switching units
Bonus: swap to advanced search and try the same thing

image

PR #530, opened 2021-02-16
Fix setting Point Radius USNG/MGRS radius

jordanwilking:
Forward port of #529


vinamartin (approved):
✅  I was seeing this, nice fix.

jlcsmith (approved):

PR #531, opened 2021-02-23
[4.1.x] Display location fields in user's preferred coordinate format

jordanwilking:
Displays the location fields in summary view and result table and list in the user's preferred coordinate format

Adds a setting to change the preferred coordinate format 

Adds WKT to the map coordinate dropdown

This pr also fixes and issue where we would sometimes try to use `.join` on non-arrays.

Testing:
1. Ingest some geo data. [Here](https://github.com/codice/ddf-ui/files/6028409/9_geo_files.zip) are some xml files with geo data.
2. Open the results and summary visualizations and show the location attribute in manage attributes
3. Select a result and, on the map, swap between coordinate settings
    image
4. Switch between results list and table and swap coordinates settings to confirm the location updates
5. In the user settings, swap between coordinate formats an verify that the formats are updated
    image
6. You can also try more complex coordinates [from here](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry).
Bonus:  add other attributes types to confirm that the `.join` issue is fixed. Adding location in manage attributes should do it, but adding a few different attributes would help.

Note: coordinates are swapped in place, so they will still appear to look wkt-like

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/package.json
+++ b/ui-frontend/packages/catalog-ui-search/package.json
@@ -172,7 +172,7 @@
     "terraformer-wkt-parser": "1.1.0",
     "urijs": "1.19.1",
     "use-resize-observer": "6.1.0",
-    "usng.js": "0.4.4",
+    "usng.js": "0.4.5",
     "uuid": "2.0.2",
     "vis": "4.15.0",
     "wellknown": "https://github.com/connexta/wellknown.git#4d9ae612e7e308509ec76400b258c9db1581ea82",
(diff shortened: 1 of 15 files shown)


vinamartin (approved):
✅  Nice!

stevenmalmgren (approved):

jlcsmith:
🎉  Hero Success 🎉 
verified that user coordinate preferences can be changed in 2D/3D map or the Map preferences area.  Verified that visuals update when the prefs change.  Confirmed that that complex geometries are correctly displayed (GeometryCollection, MultiPoint, MultiLineString, MultiPolygon, etc)

jordanwilking:
The additional code adds a fallback to UPS for MGRS when in UPS space, which has been manually tested.

PR #533, opened 2021-02-24
[4.1.x] Adds USNG/MGRS, DMS, and UTM/UPS input to Line/Poly for Search

vinamartin:
_Requires usng.js updates: https://github.com/codice/usng.js/commit/33eab1c4d383996018a9486fc6c383abd66274a8_

**What does this PR do?**
Provides DMS, USNG/MGRS, and UTM/UPS input for line and polygon options in the search pane. 
New Features:
- Degrees, Minutes, Seconds Input for line/polygon location searches
- USNG/MGRS Input for line/polygon location searches
  - Falls back to UPS for areas around poles
- UTM/UPS Input for line/polygon location searches

**How to Test**
(Use 2D map and 3D map for all of these)
- Check the Location checkbox on the search page
- Select "Line" from the dropdown
- With DD selected, click draw and draw a line on the map
- Click through the other types and make sure the line was properly translated
- Click through the other types and try typing in the points, make sure validation works, make sure typing live updates the map, make sure that you can add and delete points
- Navigate away from the search page and renavigate back and select location and line again
- For each of the types (DMS, USNG, UTM/UPS), make sure you can click the + button to add new points and type in and after you have a valid shape it appears on the map

- Repeat these steps for polygon!

Screen Shot 2021-02-24 at 14 11 54
Screen Shot 2021-02-24 at 14 12 02
Screen Shot 2021-02-24 at 14 12 36
Screen Shot 2021-02-24 at 14 12 47

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
@@ -329,12 +329,8 @@ module.exports = Backbone.AssociatedModel.extend({
 
 
   setUsngDmsUtmWithLineOrPoly(model) {
-    const key = this.get('line')
-      ? 'line'
-      : this.get('polygon')
-      ? 'polygon'
-      : undefined
-    if (key) {
+    const key = this.get('mode')
+    if (key && (key === 'line' || key === 'polygon')) {
       const points = this.get(key)
       const usngPoints = this.convertLatLonLinePolyToUsng(points)
       const dmsPoints = this.convertLatLonLinePolyToDms(points)
@@ -343,17 +339,16 @@ module.exports = Backbone.AssociatedModel.extend({
         usngPointArray: usngPoints,
         dmsPointArray: dmsPoints,
         utmUpsPointArray: utmupsPoints,
+      }, {
+        silent: true //don't trigger another onChange for line or poly
       })
+      // })
     }
   },
 
   setLatLonLinePolyFromDms() {
-    const key = this.get('line')
-    ? 'line'
-    : this.get('polygon')
-    ? 'polygon'
-    : undefined
-    if (key) {
+    const key = this.get('mode')
+    if (key && (key === 'line' || key === 'polygon')) {
       let validation = validateDmsLineOrPoly(this.get('dmsPointArray'))
       if(!validation.error) {
         const llPoints = this.get('dmsPointArray').map((point) => {
@@ -382,12 +377,8 @@ module.exports = Backbone.AssociatedModel.extend({
   },
 
   setLatLonLinePolyFromUsng() {
-    const key = this.get('line')
-    ? 'line'
-    : this.get('polygon')
-    ? 'polygon'
-    : undefined
-    if (key) {
+    const key = this.get('mode')
+    if (key && (key === 'line' || key === 'polygon')) {
       let validation = validateUsngLineOrPoly(this.get('usngPointArray'), key)
       if(!validation.error) {
         const llPoints = this.get('usngPointArray').map((point) => {
@@ -409,23 +400,20 @@ module.exports = Backbone.AssociatedModel.extend({
   },
 
   setLatLonLinePolyFromUtmUps() {
-    const key = this.get('line')
-    ? 'line'
-    : this.get('polygon')
-    ? 'polygon'
-    : undefined
-    if (key) {
+    const key = this.get('mode')
+    if (key && (key === 'line' || key === 'polygon')) {
       let validation = validateUtmUpsLineOrPoly(this.get('utmUpsPointArray'), key)
       if(!validation.error) {
         const llPoints = this.get('utmUpsPointArray').map((point) => {
-          const northPole =
-            (point.hemisphere.toUpperCase() === 'NORTHERN') ? true : false
-          const convertedPoint = converter.UTMUPStoLL({
-            northPole,
-            zoneNumber: point.zoneNumber,
-            easting: point.easting,
-            northing: point.northing,
-          })
+          // const northPole =
+          //   (point.hemisphere.toUpperCase() === 'NORTHERN') ? true : false
+          // const convertedPoint = converter.UTMUPStoLL({
+          //   northPole,
+          //   zoneNumber: point.zoneNumber,
+          //   easting: point.easting,
+          //   northing: point.northing,
+          // })
+          const convertedPoint = this.utmUpstoLL(point)
           return [
             convertedPoint.lon,
             convertedPoint.lat
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
@@ -111,21 +111,28 @@ module.exports = Backbone.AssociatedModel.extend({
       value = options
     }
     Backbone.AssociatedModel.prototype.set.call(this, key, value, options)
-    // Common.queueExecution(() => {
-    //   this.trigger('change', Object.keys(key))
-    // })
   },
 
   initialize() {
     this.listenTo(
       this,
       'change:line change:polygon',
-      this.setUsngAndDmsWithLineOrPoly
+      this.setUsngDmsUtmWithLineOrPoly
     )
     this.listenTo(
       this,
       'change:dmsPointArray',
       this.setLatLonLinePolyFromDms
+    )
+    this.listenTo(
+      this,
+      'change:usngPointArray',
+      this.setLatLonLinePolyFromUsng
+    )
+    this.listenTo(
+      this,
+      'change:utmUpsPointArray',
+      this.setLatLonLinePolyFromUtmUps
     )
     this.listenTo(
       this,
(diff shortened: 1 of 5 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
```diff
@@ -101,12 +112,21 @@ module.exports = Backbone.AssociatedModel.extend({
       value = options
     }
     Backbone.AssociatedModel.prototype.set.call(this, key, value, options)
-    Common.queueExecution(() => {
```
vinamartin, line 104: 📝  This was removed in master, see: https://github.com/codice/ddf-ui/blob/master/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js#L95-L102

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
```diff
@@ -101,12 +112,21 @@ module.exports = Backbone.AssociatedModel.extend({
...
-      this.trigger('change', Object.keys(key))
-    })
   },
 
   initialize() {
+    this.listenTo(
```
vinamartin, line 118: 📝  Setup onChange listeners for the following events:
change:line change:polygon: When a user draws on the map, update the other position formats
change:dmsPointArray: When a user manually types in DMS, update the map and other position formats 
etc.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
```diff
@@ -270,6 +290,158 @@ module.exports = Backbone.AssociatedModel.extend({
...
+      const dmsPoints = this.convertLatLonLinePolyToDms(points)
+      const utmupsPoints = this.convertLatLonLinePolyToUtm(points)
+      model.set(
+        {
+          usngPointArray: usngPoints,
+          dmsPointArray: dmsPoints,
+          utmUpsPointArray: utmupsPoints,
+        },
+        {
+          silent: true, //don't trigger another onChange for line or poly
```
vinamartin, line 347: 📝  See https://backbonejs.org/#Events-catalog for details on {silent: true}
> Generally speaking, when calling a function that emits an event (model.set, collection.add, and so on...), if you'd like to prevent the event from being triggered, you may pass {silent: true} as an option. Note that this is rarely, perhaps even never, a good idea. Passing through a specific flag in the options for your event callback to look at, and choose to ignore, will usually work out better.

The reason we need to use it here is because we are listening to the same thing we are changing.
The input for location is listening to changes from the map, but it is also changing the map. This creates a weird loop- so we want to make sure we stop the onChange trigger here.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
```diff
@@ -834,7 +1006,7 @@ module.exports = Backbone.AssociatedModel.extend({
   //
   utmUpstoLL(utmUpsParts) {
     const { hemisphere, zoneNumber, northing } = utmUpsParts
-    const northernHemisphere = hemisphere === 'NORTHERN'
+    const northernHemisphere = hemisphere.toUpperCase() === 'NORTHERN'
```
vinamartin, line 1006: 📝  Adds in a bit more safety when it comes to hemisphere case for ease of converting!

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/base.line.js
```diff
@@ -125,7 +135,6 @@ const BaseLine = (props) => {
   return (
     <div>
       <div className="input-location">
-        <Conversion value={currentValue} isValid={baseLineError} />
```
vinamartin, line 128: 📝  Gets rid of the conversion component in favor of letting the users input the usng/mgrs, dms, and utm/ups.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/dms-textfield.tsx
```diff
@@ -0,0 +1,100 @@
+/**
```
vinamartin, line 1: 📝  A wrapper for some pre-existing masked inputs for degrees, minutes, seconds.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/utmups-textfield.tsx
```diff
@@ -0,0 +1,93 @@
+/**
```
vinamartin, line 1: 📝  A wrapper for existing utm/ups input fields

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/validators.tsx
```diff
@@ -0,0 +1,229 @@
...
+}
+
+type UtmUpsPoint = {
+  easting: number
+  northing: number
+  zoneNumber: number
+  hemisphere: 'NORTHERN' | 'SOUTHERN'
+}
+
+function isUPS(input: string) {
```
vinamartin, line 45: 📝  This is a corner case validation for usng around the poles, when we get to the poles, we fall back to UPS so the textfields for usng/mgrs _may_ have UPS values. This tries to validate them by seeing if the value is deserializable as UPS which goes ahead and checks that it matches valid UPS regex.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/validators.tsx
```diff
@@ -0,0 +1,229 @@
...
+        message = errorMessages.tooFewPointsLine
+      }
+      break
+    case 'polygon':
+      if (!dms.every(validateDmsPoint)) {
+        error = true
+        message = errorMessages.invalidList
+      } else if (dms.length < 3) {
+        error = true
+        message = errorMessages.tooFewPointsPolygon
```
vinamartin, line 153: 📝  Something that's a bit curious is that it's really hard to draw a 3 point polygon. 
In almost all cases I tried manually, 3 point polys are invalid at least on the 2D map. 
However, using the 3D map it is a bit easier to get 3 point polys so I kept the existing 3 point poly error message here. 

This code is sliiiightly redundant, but I kept it expanded like this in case any of the types required extra/special validation clauses.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/dms-textfield.tsx
```diff
@@ -0,0 +1,100 @@
...
+    <div>
+      <Grid container alignItems="center">
+        <Grid item xs={11}>
+          <DmsLatitude
+            label="Latitude"
+            value={point.lat}
+            onChange={(value: string) => {
+              setPoint({
+                ...point,
+                lat: value, //validate this first
```
vinamartin: ❗ Get rid of comment

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/dms-textfield.tsx
```diff
@@ -0,0 +1,100 @@
...
+              }}
+            />
+          </DmsLatitude>
+          <DmsLongitude
+            label="Longitude"
+            value={point.lon}
+            onChange={(value: string) => {
+              setPoint({
+                ...point,
+                lon: value, //validate this first
```
vinamartin: ❗ Get rid of comment

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/base.line.js
```diff
@@ -188,4 +197,344 @@ const BaseLine = (props) => {
...
+          dmsPointArray.map((point, index) => {
+            return (
+              <div>
+                <DmsTextField
+                  key={'point-' + index}
+                  point={point}
+                  setPoint={(point) => {
+                    let array = [...dmsPointArray]
+                    array.splice(index, 1, point)
+                    setState({ ['dmsPointArray']: array })
```
vinamartin, line 234: 📝  Copying the behavior of the other geo inputs, I am setting the value as soon as a user types it whether or not its valid and then this component will validate it as the state changes. 
The reason we want to do this is so that a user can freely type without snapping and see the map get automatically updated with their points as they become valid.

vinamartin: 📝  Another note about setting the state here, this is an interesting section of code because the "state" here is a backbone model (location-old). So when this state is set, it triggers an 'onChange' on the model and when a change happens on the model, it updates the state.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
```diff
@@ -270,6 +290,158 @@ module.exports = Backbone.AssociatedModel.extend({
     }
   },
 
+  convertLatLonLinePolyToUsng(points) {
```
jordanwilking, line 293: I would name each of these to `convertLonLatLinePolyTo[type]`. Maybe even rename the points params to lonlatPoints.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
```diff
@@ -270,6 +290,158 @@ module.exports = Backbone.AssociatedModel.extend({
...
+                : 'Southern',
+          }
+        })
+      : undefined
+  },
+
+  setUsngDmsUtmWithLineOrPoly(model) {
+    let key = this.get('mode')
+    if (key === 'poly') key = 'polygon'
+    if (key && (key === 'line' || key === 'polygon')) {
```
jordanwilking, line 335: ✏️ I see that you're updating the key, so you can use it in the model.set (in the other methods), but you might factor out these bits.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/base.line.js
```diff
@@ -188,4 +197,344 @@ const BaseLine = (props) => {
...
+    if (dmsPointArray) {
+      setBaseLineError(validateDmsLineOrPoly(dmsPointArray, geometryKey))
+    }
+  }, [props.polygon, props.line, dmsPointArray])
+
+  return (
+    <div>
+      <div className="input-location">
+        {dmsPointArray &&
+          dmsPointArray.map((point, index) => {
```
jordanwilking, line 225: ✏️ I believe you can use optional chaining here `dmsPointArray?.map(() => do stuff)`.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/base.line.js
```diff
@@ -18,10 +18,20 @@ import {
   validateGeo,
```
jordanwilking, line 18: ✏️ Might just be preference, but I would create a line folder, put this in it, and factor out all of the individual components so this file is more searchable and readable.

jordanwilking (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/base.line.js
vinamartin: Oh you are right, I blanked on the syntax for this, good call 👍

stevenmalmgren (approved):
Not gonna pretend to know what's going on with some of the conversions here - everything else looks good. Nice job!

jlcsmith:
🎉  Hero Success 🎉 
Verified
- manual entry of coordinates for linestrings and polygons
- drawing on 2D/3D map and editing of UTM/MGRS/DMS coordinates for polys and lines
- verified translation of coordinates between coordinate formats

PR #534, opened 2021-02-25
[4.1.x] Update set to use filterBuilderClass for saved searches

stevenmalmgren:
Whenever set was called we didn't wrap the update to the filterTree in a filterbuilderclass which leads to the incorrect cql being parsed from the filterTree. 

Hero:
Save Searches with NOT criteria (both Field and Group level) and ensure they are preserved when loading a search.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
@@ -124,7 +124,7 @@ Query.Model = Backbone.AssociatedModel.extend({
     ) {
       // for backwards compatability
       try {
-        data.filterTree = JSON.parse(data.filterTree)
+        data.filterTree = new FilterBuilderClass(JSON.parse(data.filterTree))
       } catch (e) {
         data.filterTree = CQLUtils.transformCQLToFilter(data.cql)
       }
@@ -186,6 +186,9 @@ Query.Model = Backbone.AssociatedModel.extend({
     _.bindAll.apply(_, [this].concat(_.functions(this))) // underscore bindAll does not take array arg
     this.set('id', this.getId())
     const filterTree = this.get('filterTree')
+    if (filterTree && typeof filterTree === 'string') {
+      this.set('filterTree', JSON.parse(filterTree))
+    }
     // when we make drastic changes to filter tree it will be necessary to fall back to cql and reconstruct a filter tree that's compatible
     if (!filterTree || filterTree.id === undefined) {
       this.set('filterTree', cql.read(this.get('cql'))) // reconstruct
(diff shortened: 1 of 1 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
```diff
@@ -124,7 +124,7 @@ Query.Model = Backbone.AssociatedModel.extend({
     ) {
       // for backwards compatability
       try {
-        data.filterTree = JSON.parse(data.filterTree)
+        data.filterTree = new FilterBuilderClass(JSON.parse(data.filterTree))
       } catch (e) {
         data.filterTree = CQLUtils.transformCQLToFilter(data.cql)
```
stevenmalmgren: This isn't a function anymore. Think we should remove this.

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
```diff
@@ -124,7 +124,7 @@ Query.Model = Backbone.AssociatedModel.extend({
     ) {
       // for backwards compatability
       try {
-        data.filterTree = JSON.parse(data.filterTree)
+        data.filterTree = new FilterBuilderClass(JSON.parse(data.filterTree))
```
stevenmalmgren: Looks like this is a wrapper for anytime we call set - so it ended up with this weird feedback loop where it would overwrite the filterTree created this way in the constructor with just the JSON.parse of data.filterTree.

vinamartin (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
```diff
@@ -123,11 +123,7 @@ Query.Model = Backbone.AssociatedModel.extend({
       typeof data.filterTree === 'string'
     ) {
       // for backwards compatability
-      try {
-        data.filterTree = JSON.parse(data.filterTree)
-      } catch (e) {
-        data.filterTree = CQLUtils.transformCQLToFilter(data.cql)
-      }
+      data.filterTree = new FilterBuilderClass(JSON.parse(data.filterTree))
```
jordanwilking, line 126: ❓ Should we be concerned with the backwards compatibility with cql that's mentioned here?

jordanwilking (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Query.js
stevenmalmgren: @andrewkfiedler would know. I don't believe so. My understanding is that the filterTree being passed in, not as a FilterBuilderClass, is the backwards compatibility we're concerned with - so this is necessary. I think the issue stems from the assumption the last line in the constructor would add the FilterBuilderClass, but since set is called this method ends up having the final say.

PR #540, opened 2021-03-08
[4.1.x] Adds storeId to validation endpoint

vinamartin:
Adds storeId to validation endpoint for items in another catalog store that need to be validated. 

To test:
Set up another local store (source) and have an item come from that source (e.g. Harvested source).
Then verify the quality tab doesn't show a giant yellow error saying the endpoint has errored out.

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
@@ -276,7 +276,21 @@ public class MetacardApplication implements SparkApplication {
           String id = req.params(":id");
           return util.getJson(validator.getFullValidation(util.getMetacardById(id)));
         });
+    get(
+        "/metacard/:id/:storeId/attribute/validation",
+        (req, res) -> {
+          String id = req.params(":id");
+          String storeId = req.params(":storeId");
+          return util.getJson(validator.getValidation(util.getMetacardById(id, storeId)));
+        });
 
+    get(
+        "/metacard/:id/:storeId/validation",
+        (req, res) -> {
+          String id = req.params(":id");
+          String storeId = req.params(":storeId");
+          return util.getJson(validator.getFullValidation(util.getMetacardById(id, storeId)));
+        });
     post(
         "/prevalidate",
         APPLICATION_JSON,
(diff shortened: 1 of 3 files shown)


ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -276,7 +276,21 @@ public void init() {
           String id = req.params(":id");
           return util.getJson(validator.getFullValidation(util.getMetacardById(id)));
         });
+    get(
```
vinamartin, line 279: ✏️  I added these as new lines just in case we need the validation endpoint without the storeId. Not sure if it's better to collapse them.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/metacard-quality/container.tsx
```diff
@@ -46,13 +46,14 @@ class MetacardQuality extends React.Component<Props, State> {
   componentDidMount() {
     setTimeout(() => {
       const metacardId = this.model.get('metacard').get('id')
+      const storeId = this.model.get('metacard').get('properties').get('source-id')
 
       const attributeValidationRes = fetch(
-        `./internal/metacard/${metacardId}/attribute/validation`
+        `./internal/metacard/${metacardId}/${storeId}/attribute/validation`
```
vinamartin, line 55: ❓ Anywhere that's hitting this endpoint now will cause an error if the item is not in the local catalog. Should those other places be updated?

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/CqlQueriesImpl.java
```diff
@@ -215,7 +215,7 @@ public CqlQueryResponse executeCqlQuery(CqlRequest cqlRequest)
    * @param request Catalog Query Request
    * @param responses List of responses to append to.
    * @return A ResultIterable of results, additionally adding the query response to a mutatable list
-   *     for additional context as we query.
+   *     for additiol context as we query.
```
stevenmalmgren, line 218: love this

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/metacard-quality/container.tsx
stevenmalmgren: Yea, generally a proponent of being explicit about the source even if it is local. The implicit assumption that it's local tends to burn us.

vinamartin:
❗ This appears to create weird errors on upload.

stevenmalmgren (approved):

jordanwilking (approved):

stevenmalmgren:
### **🥇HERO SUCCESSFUL🥇**
Set up harvested source and ingested resource from said source. Quality tab would not error out.

PR #543, opened 2021-03-12
Fix NPE when highlighting missing attribute

derekwilhelm:
When highlights are returned for attributes that have been removed from the resulting metacard or don't exist, an NPE would be thrown.  This PR adds a check to make sure the attribute and referenced value exists before processing the highlight.

### DIFF
--- a/ui-backend/catalog-plugin-highlight/src/main/java/org/codice/ddf/catalog/ui/plugin/highlight/HighlightTransformPlugin.java
+++ b/ui-backend/catalog-plugin-highlight/src/main/java/org/codice/ddf/catalog/ui/plugin/highlight/HighlightTransformPlugin.java
@@ -27,15 +27,20 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 /** Transforms solr highlights into an easily displayable format on the frontend */
 public class HighlightTransformPlugin implements PostQueryPlugin {
 
   private int bufferSize;
 
+  private Pattern redactedPattern;
+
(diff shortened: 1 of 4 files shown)


jordanwilking (approved):

ui-backend/catalog-plugin-highlight/src/main/java/org/codice/ddf/catalog/ui/plugin/highlight/HighlightTransformPlugin.java
```diff
@@ -116,9 +118,12 @@ private void processHighlight(
...
       Highlight highlight) {
     Attribute attribute = matchingResult.getMetacard().getAttribute(attributeName);
-    String value = (String) attribute.getValues().get(highlight.getValueIndex());
+    String value = null;
+    if (attribute != null && !attribute.getValues().isEmpty()) {
+      value = (String) attribute.getValues().get(highlight.getValueIndex());
+    }
 
-    if (!value.equals("REDACTED")) {
+    if (value != null && !value.equals("REDACTED")) {
```
greg-fraser: If we add the per Source Point of Contact information (ISR-7159) then this will contain more than "REDACTED".
Would using `value.contains()` or `value.startsWith()` make more sense?
Also would be safer if we are referring to a common constant for the string "REDACTED" - not sure where this would live?

vinamartin (approved):

derekwilhelm:
build now

jordanwilking (approved):

stevenmalmgren:
### **🥇HERO SUCCESSFUL🥇**
Verified fields that match the redacted criteria are no longer included in highlights for single value and multi valued fields.

PR #544, opened 2021-03-18
[4.1.x] Adds exception to escaping for the id property

andrewkfiedler:
- Adds exception in both directions (write and read)
     - Adds tests for each scenario, with verification that other fields still escape while id does not.

To test manually you can look at the outgoing request difference between an id query and another property, such as seen in these screenshots.

Screen Shot 2021-03-17 at 11 23 17 PM
Screen Shot 2021-03-17 at 11 22 57 PM

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/cql.spec.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/cql.spec.tsx
@@ -13,6 +13,10 @@
  *
  **/
 import { expect } from 'chai'
+import {
+  FilterBuilderClass,
+  FilterClass,
+} from '../component/filter-builder/filter.structure'
 import cql from './cql'
 
 type CapabilityCategoriesType =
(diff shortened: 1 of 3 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/js/cql.spec.tsx
```diff
@@ -161,4 +165,48 @@ describe('read & write parity for capabilities, as well as boolean logic', () =>
...
+      const newFilter = cql.read(cqlText)
+      const filterToCheck = newFilter.filters[0] as FilterClass
+      expect(cqlText, `Doesn't escape properly`).to.equal(
+        '("title" = \'12123123\\_123213123\')'
+      )
+      expect(filterToCheck.value).to.equal(value)
+    })
+
+    it('it handles escaping _ in properties that are id', () => {
+      const value = '12123123_123213123'
```
glenhein, line 192: What about the case where the value is `12123123\_123213123`

glenhein (approved):

derekwilhelm (approved):

jordanwilking (approved):

jordanwilking:
🌐  Hero Successful  🌐 

#### Steps

1.  Disabled websockets
2. Confirmed that the escape character wasn't added to the id field
image

PR #545, opened 2021-03-18
Adds exception to escaping for the id property

andrewkfiedler:
- Adds exception in both directions (write and read)
     - Adds tests for each scenario, with verification that other fields still escape while id does not.

To test manually you can look at the outgoing request difference between an id query and another property, such as seen in these screenshots.

Screen Shot 2021-03-17 at 11 23 17 PM
Screen Shot 2021-03-17 at 11 22 57 PM

Forward port of https://github.com/codice/ddf-ui/pull/544


glenhein (approved):

derekwilhelm (approved):

jordanwilking (approved):

jordanwilking:
Heroed in https://github.com/codice/ddf-ui/pull/544

PR #547, opened 2021-03-23
[4.2.x] Adds protection against missing model or empty lists for expand metacard interaction

andrewkfiedler:
- Adds tests that exercise a missing model, as well as an empty list being passed.
 - This would happen if the multiselect menu happened to be open, and then the selection changed to be either undefined or empty suddenly (select two, open menu, while open deselect everything).

https://user-images.githubusercontent.com/11984853/112079621-bb812100-8b3d-11eb-845a-9edda8a0ce51.mov

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/cql.spec.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/cql.spec.tsx
@@ -13,6 +13,10 @@
  *
  **/
 import { expect } from 'chai'
+import {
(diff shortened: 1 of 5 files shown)


jordanwilking (approved):

samuelechu (approved):

jordanwilking:
🌐  Hero Successful  🌐 

#### Steps

1. Verified that I could reproduce the problem (so I could tell if it is fixed).
2. Selected multiple resources, opened the selected dropdown, and deselected everything. UI didn't break 🎉

PR #551, opened 2021-03-23
apply attribute aliases to built-in attributes

glenhein:
This PR applies the configurable attribute aliases to the built-in attributes, such as anyText and anyGeo.


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/singletons/metacard-definitions.js
```diff
@@ -49,41 +49,49 @@ function transformEnumResponse(metacardTypes, response) {
     {}
   )
 }
+
 const metacardStartingTypes = {
   anyText: {
+    alias: properties.attributeAliases['anyText'],
```
glenhein, line 55: Is there a cleaner way to do this?

PR #552, opened 2021-03-23
[master] apply attribute aliases to built-in attributes (eg anyText)

glenhein:
This PR applies the configurable attribute aliases to the built-in attributes, such as anyText and anyGeo.


samuelechu (approved):

jordanwilking:
Heroed in #553

PR #558, opened 2021-03-29
Fixes datepickers not accounting for DST correctly, turns on tests in CI

andrewkfiedler:
- Turn on tests in CI
 - Adds tests / fix for daylight savings time offset
 - Fixes DST issue: We need to pass in the date to the offset calculation for moment, otherwise it won't take into account the daylight savings time shift as well for that timezone.
 - Swap build / test phases (The tgz generated from the build phase will be accidentally deleted if we don't.)

Screenshots:  
Screen Shot 2021-03-24 at 2 53 10 PM
 - Notice it's off by one hour from the selection in the datepicker.

Screen Shot 2021-03-24 at 2 53 57 PM

Testing:
Verify that going across the dateline in either direction, the datepicker is always in sync with what is in the text box.  If you want, you can change you computer's time pref to test it more thoroughly.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-helpers.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-helpers.tsx
@@ -70,7 +70,7 @@ export const DateHelpers = {
     DateProps: {
       generateOnChange: (onChange: (value: string) => void) => {
         return ((selectedDate, isUserChange) => {
-          if (onChange && selectedDate && isUserChange) {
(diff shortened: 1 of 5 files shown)


andrewkfiedler:
build now

andrewkfiedler:
Checking out the CI issue, it looks like the tests caught something 🎉

andrewkfiedler:
build now

stevenmalmgren (approved):

jordanwilking (approved):

andrewkfiedler:
build now

PR #568, opened 2021-04-07
[4.2.x] Remove frontend usage of federation, updates query to only use user defaults if flag is passed

andrewkfiedler:
- Updates query to by default not use the user specified settings.
 - Removes code paths that are no longer needed.

Testing:  
 - Update your user settings to have a selected set of sources (such as slow).
 - Verify creating a new search will utilize those sources.
 - Verify that when visiting the id page of a metacard, the query that goes out does not use the users setting for sources (set it to something that the metacard isn't in, or just look at the network).


andrewkfiedler:
I just realized I need to do a similar thing for the ephemeral filtering and sorting (they are currently applied in scenarios they should not be, ex: add a filter after searching, then go to an id page you should be able to see and it will be inadvertently filtered).  I should be able to add a similar fix to the sources fix.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/metacard.tsx
```diff
@@ -3,33 +3,57 @@ import { hot } from 'react-hot-loader'
...
+            })
+          }
+        })
+        .concat(
+          new FilterClass({
+            type: '=',
+            value: '-1',
+            property: 'id',
+          })
+        )
```
jordanwilking, line 55: ❓ So, if the file doesn't have any children, we default a filter with the file id and these lines. What does this do?

jordanwilking (approved):
Liking the hooks 👍

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/metacard.tsx
andrewkfiedler: I'm trying to remember why exactly, but it's been so long I can't.  From what I recall the queries didn't finish correctly without this.

However, at this point it might be a left over of an old limitation that is no longer necessary.

jordanwilking:
ℹ️ Hero Successful ℹ️ 

#### Steps

1. Changed user settings to slow(offsite) in Settings -> Search Options
2. Navigated to search and verified that it had my search
    image
3. Ingested a file and went to its id page. Verified that the network didn't use the user settings
    image
4. Removed all sources in settings and tried again to verify that the id query was unaffected.
5. Added a harvested source, set that as my source in settings, and tried again. The metacard Id route didn't use the user settings sources.
    image

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/home.tsx
```diff
@@ -206,24 +206,24 @@ const LeftBottom = ({ selectionInterface }: { selectionInterface: any }) => {
...
-  const [queryModel, setQueryModel] = React.useState(
-    urlBasedQuery || new Query.Model()
-  )
+}
+
+export const HomePage = () => {
+  const location = useLocation()
+  let urlBasedQuery = location.search.split('?defaultQuery=')[1]
+  const [queryModel] = useUserQuery(decodeUrlIfValid(urlBasedQuery))
+  console.log(queryModel)
```
samuelechu: remove?

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/source-selector.tsx
```diff
@@ -87,27 +87,13 @@ const shouldBeSelected = ({
  * If it includes 'remote', that that means everything remote.  All other values are singular selections of a source.
```
samuelechu, line 87: Does this type need to be updated to remove 'enterprise'? https://github.com/codice/ddf-ui/blob/7fbc7060ccceb0f98184eb59942a742a999c8bab/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/source-selector.tsx#L21

samuelechu (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-settings/source-selector.tsx
andrewkfiedler: Sort of, it should be 'all' 'fast' 'slow' string now

PR #570, opened 2021-04-12
[4.2.x] 2D Map view zooms in when not active tab

jordanwilking:
#### Description

When the map view isn't the active tab, and we calculate its size (e.g. when running a search), it gets a bad zoom level and zooms all the way in. This adds the same zoom level as the other places where we zoom in.

#### Testing

1. Navigate to Search and open the 2D map and Inspector view - "tabbed" together with the Inspector view as the active view
2. Run the search and switch the 2D map tab
3. Verify that the 2D map isn't zoomed all the way in (example below)


jlcsmith (approved):

pklinef (approved):

andrewkfiedler (approved):

samuelechu:
Hero successful! Verified that the 2d map isn't zoomed in all the way after performing a search where the 2d map tab is not active.

PR #571, opened 2021-04-12
2D Map view zooms in when not active tab

jordanwilking:
Forward port for #570


pklinef (approved):

samuelechu (approved):

andrewkfiedler (approved):

samuelechu:
Hero successful! Verified that the 2d map isn't zoomed in all the way after performing a search where the 2d map tab is not active.

jordanwilking:
build now

PR #573, opened 2021-04-15
[4.2.x] Added AnyText and AnyGeo special sort case for Search Attribute

jMoneee:
Description
Adds a special case in the sorting of the list of search attributes that puts anyText and anyGeo at the top no matter the alias given.

Testing
1. Build and Install UI
2. Add aliases for anyTest and AnyGeo in the Admin UI under System>Catalog UI Search Attribute Aliases
-- anyText=Search - Any Text
-- anyGeo=Search - Any Geo
3. Verify that the new alias appears at the top of the list of search attributes in the advanced search menu

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/singletons/metacard-definitions.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/singletons/metacard-definitions.js
@@ -227,6 +227,10 @@ module.exports = new (Backbone.Model.extend({
     return metacardTypes.sort((a, b) => {
       const attrToCompareA = (a.alias || a.id).toLowerCase()
       const attrToCompareB = (b.alias || b.id).toLowerCase()
+      if (a.id == 'anyText') return -1
+      if (b.id == 'anyText') return 1
+      if (a.id == 'anyGeo') return -1
+      if (b.id == 'anyGeo') return 1
       if (attrToCompareA < attrToCompareB) {
         return -1
       }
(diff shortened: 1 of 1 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/singletons/metacard-definitions.js
```diff
@@ -227,6 +227,10 @@ module.exports = new (Backbone.Model.extend({
     return metacardTypes.sort((a, b) => {
       const attrToCompareA = (a.alias || a.id).toLowerCase()
       const attrToCompareB = (b.alias || b.id).toLowerCase()
+      if (a.id == 'anyText') return -1
```
jordanwilking: ✏️ Just a suggestion/option, but it might be a bit easier to add new values (or eventually make an admin config) if we do something like:

```
const PRIORITY_ATTRIBUTES = ["anyText", "anyGeo"] // or some better name - outside the function

if (PRIORITY_ATTRIBUTES.includes(a.id)) return -1
if (PRIORITY_ATTRIBUTES.includes(b.id)) return 1
```

jordanwilking (approved):

samuelechu (approved):

andrewkfiedler (approved):

jordanwilking:
👽 Hero Successful 👽 

#### Steps

1. Added the aliases in the admin config 
    image
2. Confirmed that the anyText and anyGeo aliases end up at the top of the list of searchable attributes.

PR #572, opened 2021-04-13
[WIP] [4.2.x] Update result refresh

samuelechu:
Adds an option to specify metacardProperties when refreshing a result metacard. This is necessary due to solr occasionally not providing the latest version of data.

PR #574, opened 2021-04-19
[4.2.x] Update result refresh

samuelechu:
Adds an option to specify metacardProperties when refreshing a result metacard. This is necessary due to solr occasionally not providing the latest version of data. This code is the same as https://github.com/codice/ddf-ui/pull/572, except we skip the network call (since it may provide outdated data) if metacard properties are passed through and instead update the current model with the provided metacard properties.

Testing: Verify that the behavior of the history, archive, and overwrite tabs of the inspector view behave the same as before (These are the views that use the `refreshData` function).

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/ResultUtils.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/ResultUtils.js
@@ -16,9 +16,9 @@ const _ = require('underscore')
 const metacardDefinitions = require('../component/singletons/metacard-definitions.js')
 
 module.exports = {
-  refreshResult(result) {
+  refreshResult(result, metacardProperties) {
     const id = result.get('metacard').id
-    result.refreshData()
+    result.refreshData(metacardProperties)
   },
   updateResults(results, response) {
     const attributeMap = response.reduce(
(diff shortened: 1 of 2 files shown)


jordanwilking (approved):

andrewkfiedler (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/ResultUtils.js
```diff
@@ -16,9 +16,9 @@ const _ = require('underscore')
 const metacardDefinitions = require('../component/singletons/metacard-definitions.js')
 
 module.exports = {
-  refreshResult(result) {
+  refreshResult(result, metacardProperties) {
     const id = result.get('metacard').id
```
andrewkfiedler, line 20: I know this is outside your changes, but why do we do this id declaration?  🤔

samuelechu: was wondering this as well :shrug:

jordanwilking:
🦘 Hero Successful 🦘 

#### Steps

1.  Ingested resources and changed the title of one a few times to create some history
2. Performed a few reversions and the both the Summary and Results views updated as expected. Refreshed to confirm that the changes were being saved.
3. Confirmed that I was able to overwrite a resource with another new one.
4. Successfully archived a resource
image

PR #575, opened 2021-04-19
[4.2.x] Fix source property collision

pklinef:
Description
Highlights from multiple sources can collide since they all use the same source response property key.
Now uses the new per source response properties if available to find all source highlights to avoid missing highlights due to collisions.

Related to https://github.com/codice/ddf/pull/6560


jordanwilking (approved):

derekwilhelm (approved):

jlcsmith:
@pklinef  - this depends on the next DDF release (2.26.9) correct?

pklinef:
To get the fixed behavior, it requires the changes from codice/ddf#6560. But it is backwards compatible so it falls back to its old behavior without that other change.

derekwilhelm:
🎉 Hero Successful 🎉 

Built this with the DDF changes and tested with multiple sources that provide highlight data.  The data was properly separated out into original source properties and were aggregated together to send to the UI so that all of the highlight data was visible.

derekwilhelm:
build now

jordanwilking:
Merging because this one is backwards compatible.

PR #576, opened 2021-04-21
[4.2.x] Add structure for custom metacard update handling for summary view

jordanwilking:
Downstream should handle this a bit differently. This adds the structure for an extension point. It also factors out the original method to make the extension swap a bit more understandable.

### Testing

1. Upload a resource
2. Navigate to search and search for that resource.
3. Select it and update an attribute (e.g. title) in the summary view (in the Inspector visualization)
4. Verify that the update works. This can be checked by making sure the attribute is updated - and refreshing to double check. And/or making sure the networkd request going out is going to the `metacards?storeId` endpoint

Note: this could be tested with downstream, but it's a bit more work.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
@@ -37,6 +37,8 @@ import DateTimePicker from '../../fields/date-time-picker'
 import Geometry from '../../../react-component/input-wrappers/geometry'
 import { useRerenderOnBackboneSync } from '../../../js/model/LazyQueryResult/hooks'
 import useCoordinateFormat from './useCoordinateFormat'
+import { MetacardAttribute } from '../../../js/model/Types'
+import ExtensionPoints from '../../../extension-points'
 
 function getSummaryShown(): string[] {
   const userchoices = user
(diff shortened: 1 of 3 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
```diff
@@ -360,34 +387,24 @@ export const Editor = ({
...
+              setTimeout(() => {
+                addSnack('Successfully updated.')
+                onSave()
+              }, 1000)
+
+            if (ExtensionPoints.handleMetacardUpdate) {
+              ExtensionPoints.handleMetacardUpdate(
+                lazyResult,
+                attributes,
+                onComplete
```
samuelechu: Just a thought (don't know if this will work or if it's any better than what you have currently), but maybe we could have the handleMetacardUpdate function return a promise and call the onComplete afterwards upstream instead?

samuelechu (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
jordanwilking: That's a good thought. I'll take a look here in a bit.

jordanwilking:
Can't seem to push to the branch now for some reason. Need to add a small change before merging.

glenhein (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
```diff
@@ -137,6 +139,31 @@ enum Mode {
   BadInput = 'bad-input',
 }
 
+const handleMetacardUpdate = (
```
andrewkfiedler: I like to stick to one parameter, but send things in as an object.  I find it simpler to maintain.

andrewkfiedler (approved):

andrewkfiedler:
Verified ✅ 
Uploaded an image.  Verified that changing things in the inspector worked appropriately and the "successfully updated snack fired".

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
```diff
@@ -137,6 +139,35 @@ enum Mode {
...
+    $.ajax({
+      url: `./internal/metacards?storeId=${lazyResult.plain.metacard.properties['source-id']}`,
+      type: 'PATCH',
+      data: JSON.stringify(payload),
+      contentType: 'application/json',
+    })
+      .then((response: any) => {
+        ResultUtils.updateResults(lazyResult.getBackbone(), response)
+      })
+      .always(onComplete)
```
andrewkfiedler, line 167: The one thing I realized is odd (we should make a ticket to go back and address), is that since we do an "always" here it doesn't matter if the update succeeds or not.  We should update this to have an onSuccess, onFailure, onComplete so that we can respond appropriately.  Right now if a failure occurs it'll still say "successfully updated".

PR #578, opened 2021-04-22
[4.2.x] Update query status dropdown

samuelechu:
Adds scrolling to the query status dropdown.
Screenshot from 2021-04-22 12-29-16


jordanwilking (approved):

jordanwilking:
ℹ️ Hero Successful ℹ️ 

#### Notes

We may want to set a max height on the popover, so it's a bit less intrusive when there are a large number of sources.

#### Steps

1. Added a few dummy harvested sources in the admin ui
2. In query-feed.tsx, I duplicated the row rendering logic to give me extra sources
3. Verified that the popover becomes scrollable when the sources would overflow
image
image

andrewkfiedler:
Verified: 
Screen Shot 2021-04-25 at 11 17 38 PM

andrewkfiedler (approved):

PR #577, opened 2021-04-22
BR-169 Upgrade ddf-ui to use jdk11

TonyMorrison:
## What does this PR do?
Modifies the Jenkinsfile to use jdk11 in the maven steps of the build. The rest of the code has already been upgraded to use Java 11

## Reviewers
@shaundmorris 
@AzGoalie

PR #579, opened 2021-04-22
Update result refresh

samuelechu:
Forward port of https://github.com/codice/ddf-ui/pull/574

Adds an option to specify metacardProperties when refreshing a result metacard. This is necessary due to solr occasionally not providing the latest version of data. This code is the same as #572, except we skip the network call (since it may provide outdated data) if metacard properties are passed through and instead update the current model with the provided metacard properties.

Testing: Verify that the behavior of the history, archive, and overwrite tabs of the inspector view behave the same as before (These are the views that use the refreshData function).

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/ResultUtils.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/ResultUtils.js
@@ -16,9 +16,8 @@ const _ = require('underscore')
 const metacardDefinitions = require('../component/singletons/metacard-definitions.js')
 
 module.exports = {
-  refreshResult(result) {
-    const id = result.get('metacard').id
-    result.refreshData()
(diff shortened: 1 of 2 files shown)


jordanwilking:
build now

jordanwilking:
Build now, successfully. Please.

samuelechu:
I think we might need https://github.com/codice/ddf-ui/pull/577 to be merged due to the upgrade to java 11 :sweat_smile:

jordanwilking:
> I think we might need #577 to be merged due to the upgrade to java 11 😅

That probably makes sense 😂. At least we have the 4.2.x one in.

shaundmorris:
build now

LinkMJB:
build now

LinkMJB:
build now

samuelechu:
build now

jordanwilking:
Tested in #574

PR #586, opened 2021-05-04
USNG Bbox drawing doesn't update the input values in search

samuelechu:
image

The silent option passed to the change event after drawing when the usng tab is active was introduced to address an issue where a loop occurred when switching to the USNG tab of the location component from the lat/lon tab. 
https://github.com/codice/ddf/commit/3fc35872ebef39feb77f3031c5e82692a2dc4c97
https://github.com/codice/ddf/commit/3fc35872ebef39feb77f3031c5e82692a2dc4c97#diff-089bf8013be350ca9014bc41912b4f9e3ff12404d19fffb40fee24ee324ed14fR123
I verified that this behavior is no longer present when removing this conditional.

PR #588, opened 2021-05-04
[4.2.x] USNG Bbox drawing doesn't update the input values in search

samuelechu:
Backport of https://github.com/codice/ddf-ui/pull/586
image

The silent option passed to the change event after drawing when the usng tab is active was introduced to address an issue where a loop occurred when switching to the USNG tab of the location component from the lat/lon tab. 
https://github.com/codice/ddf/commit/3fc35872ebef39feb77f3031c5e82692a2dc4c97
https://github.com/codice/ddf/commit/3fc35872ebef39feb77f3031c5e82692a2dc4c97#diff-089bf8013be350ca9014bc41912b4f9e3ff12404d19fffb40fee24ee324ed14fR123
I verified that this behavior is no longer present when removing this conditional.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
@@ -684,10 +684,7 @@ module.exports = Backbone.AssociatedModel.extend({
       west !== undefined
     ) {
       this.set('bbox', [west, south, east, north].join(','), {
-        silent:
-          (this.get('locationType') === 'usng' ||
-            this.isLocationTypeUtmUps()) &&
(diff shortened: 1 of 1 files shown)


jMoneee (approved):

jordanwilking (approved):

jordanwilking:
🌐  Hero Successful  🌐 

#### Steps

1. Drew a bounding boxes both with USNG/MGRS selected and with a different coordinate system selected. USNG/MGRS was successfully update.
2. Verified that I was able to type into the USNG/MGRS inputs and the bounding box was updated. Was also able to subsequently update via drawing a bounding box.

image


#### Notes
  -  The USNG/MGRS input box still often display errors, but that's beyond the scope of this ticket.

PR #592, opened 2021-05-11
[4.2.x] Update table export to use transformFilterTree instead of deleted getEphemeralMixinCql

andrewkfiedler:
- Updates the table export component to no longer reference the deleted internal method of query.  Instead, it uses the transformFilterTree function in TypedQuery.

To Test:
Verify that you can export results to csv, and then verify that adding a temporary filter (above the visuals on the right) and exporting will export the correct results. 

Screen Shot 2021-05-11 at 9 27 55 AM


samuelechu (approved):

jordanwilking (approved):

jMoneee:
Hero Successful
Successfully exported correct result sets to CSV both with and without the temporary filter

PR #593, opened 2021-05-12
Add extension point in revert so downstream projects can supply custom catalog properties

glenhein:
A downstream project needs to set some extra properties for catalog query and catalog create operations during reverts. This PR adds an extension point to supply those properties. The solution is implemented so that it should be easy to expand its use to other parts of ddf-ui.


ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -403,7 +408,7 @@ public void init() {
           String revertId = req.params(":revertid");
           String storeId = req.params(":storeId");
           Metacard versionMetacard = revert(id, revertId, storeId);
-          return util.metacardToJson(MetacardVersionImpl.toMetacard(versionMetacard, types));
```
glenhein, line 406: Just removing an argument that is no longer used. The old form of toMetacard is deprecated and the "types" argument is now ignored.

ui-backend/catalog-ui-search/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -253,6 +253,12 @@ Implementation details
         <argument ref="catalogFramework"/>
     </bean>
 
+    <reference-list id="revertProperties" interface="org.codice.ddf.catalog.ui.metacard.internal.PropertySupplier" filter="(type=revert)"/>
```
glenhein: I'm using a filter to establish a convention on how downstream projects can supply properties for other aspects of ddf-ui.

ui-backend/catalog-ui-search/ou
```diff
@@ -0,0 +1,208 @@
+[INFO] Scanning for projects...
```
pklinef: ❗ Remove this file.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/internal/PropertySupplier.java
```diff
@@ -0,0 +1,27 @@
...
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ddf.catalog.ui.metacard.internal;
+
+import java.io.Serializable;
+import java.util.Map;
+
+/** Supply addititonal properties for catalog operations. This allows properties to be injected. */
+public interface PropertySupplier {
```
pklinef: ✏️ I would suggest renaming this to `OperationPropertySupplier` to make it clearer which properties this applies to.

https://github.com/codice/ddf/blob/master/catalog/core/catalog-core-api/src/main/java/ddf/catalog/operation/Operation.java#L31

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/internal/PropertySupplier.java
```diff
@@ -0,0 +1,27 @@
...
+
+  /** Return properties to be added to query operations. */
+  Map<String, Serializable> queryProperties();
```
pklinef: ❗ There are many other types of operation properties like update, delete, resource, storage, and source info. It might be a good idea to make this more generic so we don't need to add a new method for each new implementation of `Operation`.

pklinef (changes requested):

ui-backend/catalog-ui-search/ou
glenhein: oops

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/internal/PropertySupplier.java
pklinef: One suggestion would be a `getProperties` method that takes the  class that implements `Operation` of interest as a lookup key.

glenhein: I went with a string parameter that indicates the type of operation.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/internal/OperationPropertySupplier.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package org.codice.ddf.catalog.ui.metacard.internal;
+
+import java.io.Serializable;
+import java.util.Map;
+
+/** Supply addititonal properties for catalog operations. This allows properties to be injected. */
+public interface OperationPropertySupplier {
+
+  String QUERY_TYPE = "query";
```
pklinef, line 22: ✏️ I would remove these string constants from this interface. Otherwise developers might be tempted to included strings related to operations all over the system or even external repositories. That happened with `PersistentStore`. For example, these could probably live in `MetacardApplication` instead. Ideally this interface should never change just because a new implementation of this interface is created for a new operation type.

pklinef (approved):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/internal/OperationPropertySupplier.java
glenhein: For now, I'm going to leave the strings. Hopefully, I don't live to regret it ;-)

jlcsmith (approved):

glenhein:
Superseded by https://github.com/codice/ddf-ui/pull/598

PR #598, opened 2021-05-18
Add extension point in revert so downstream projects can supply custom catalog properties

glenhein:
A downstream project needs to set some extra properties for catalog query and catalog create operations during reverts. This PR adds an extension point to supply those properties. The solution is implemented so that it should be easy to expand its use to other parts of ddf-ui.

Testing Instructions: Make sure reverts are working as expected.

Note: this pr replaces https://github.com/codice/ddf-ui/pull/593

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/internal/OperationPropertySupplier.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/internal/OperationPropertySupplier.java
@@ -1,0 +1,31 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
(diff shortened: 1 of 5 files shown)


jlcsmith (approved):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/internal/OperationPropertySupplier.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.io.Serializable;
+import java.util.Map;
+
+/** Supply addititonal properties for catalog operations. This allows properties to be injected. */
+public interface OperationPropertySupplier {
+
+  String QUERY_TYPE = "query";
+
+  String CREATE_TYPE = "create";
```
derekwilhelm, line 24: ❓ Should updates and deletes have this capability as well?

derekwilhelm (approved):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/internal/OperationPropertySupplier.java
glenhein: Eventually, if/when we need them. But for now, I was going to limit it just what is needed.

jlcsmith:
🎉  Hero Success 🎉 
Reverts work as expected

jlcsmith:
build now

PR #600, opened 2021-06-02
increase websocket max text message size

glenhein:
A downstream project was experiencing trouble with searches using very large geometries. It was found that increasing the max text message size for web sockets fixed the problem. I decided to not make this configurable, but I could if the team believes it would be useful. I'm trying to not add yet-another-config.

Test by making sure searches using large geometries are working as expected.

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/ws/JsonRpc.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/ws/JsonRpc.java
@@ -45,6 +45,8 @@ public class JsonRpc implements WebSocket {
   private static final String METHOD = "method";
   private static final String ID = "id";
 
+  private static final int WS_MAX_TEXT_MSG_SIZE = 1_000_000;
+
   private static final Gson GSON =
       new GsonBuilder()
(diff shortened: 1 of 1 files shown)


ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/ws/JsonRpc.java
```diff
@@ -96,6 +98,7 @@ private static Error invalid(String message, Object data) {
   @Override
   public void onOpen(Session session) {
     // no action required on open
```
glenhein: I probably should remove this comment, since it is no longer accurate

jlcsmith (approved):

jlcsmith:
You can use Luxembourg waterways - a public shapefile from ArcGIS https://mapcruzin.com/download-shapefile/luxembourg-waterways-shape.zip

jlcsmith:
🎉  Hero success 🎉 

Verified that the websocket message size change does fix the issue with large / complex search geometries.

jlcsmith:
build now

jordanwilking (approved):

derekwilhelm (approved):

PR #603, opened 2021-06-14
Update location-old to handle point radius init

jordanwilking:
As far as I can tell, initializing a location-old model doesn't propagate the initial values to the other coordinate types. So, if we pass lat and lon to it, utmpUps and usng won't be updated.

This enables propagation to other coordinate types when location-old is passed a point radius, which appears to be our only usecase for initializing location-old models with values.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
@@ -112,8 +112,7 @@ module.exports = Backbone.AssociatedModel.extend({
     }
     Backbone.AssociatedModel.prototype.set.call(this, key, value, options)
   },
-
-  initialize() {
+  initialize(props) {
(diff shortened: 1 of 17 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
```diff
@@ -566,12 +575,7 @@ module.exports = Backbone.AssociatedModel.extend({
     const lat = this.get('lat'),
       lon = this.get('lon')
 
-    if (
-      (!Drawing.isDrawing() && this.get('locationType') !== 'latlon') ||
-      !this.isLatLonValid(lat, lon)
-    ) {
-      return
-    }
+    if (!this.isLatLonValid(lat, lon)) return
```
jordanwilking, line 577: This allows usng and utmUps to be updated for point radii outside of drawing

PR #604, opened 2021-06-15
Update location-old to handle point radius init

jordanwilking:
As far as I can tell, initializing a location-old model doesn't propagate the initial values to the other coordinate types. So, if we pass lat and lon to it, utmpUps and usng won't be updated.

This enables propagation to other coordinate types when location-old is passed a point radius, which appears to be our only usecase for initializing location-old models with values.

For testing, it may just be easiest to copy this file into node_modules with the dev server running. 

1. Ingest metacards with geo data. [articles.zip](https://github.com/codice/ddf-ui/files/6656178/articles.zip)
2. Search for the metacards and select "Create Search From Location". This will populate the Point Radius geo in the search.
3. Verify that each coordinate system is populated.
image

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
@@ -112,8 +112,7 @@ module.exports = Backbone.AssociatedModel.extend({
     }
     Backbone.AssociatedModel.prototype.set.call(this, key, value, options)
   },
-
-  initialize() {
+  initialize(props) {
     this.listenTo(
       this,
       'change:line change:polygon',
@@ -186,6 +185,15 @@ module.exports = Backbone.AssociatedModel.extend({
     })
     this.listenTo(this, 'EndExtent', this.drawingOff)
     this.listenTo(this, 'BeginExtent', this.drawingOn)
+    this.initializeValues(props)
+  },
+  initializeValues(props) {
+    if (props.type === 'POINTRADIUS' && props.lat && props.lon) {
+      if (!props.usng || !props.utmUpsEasting) {
+        // initializes dms/usng/utmUps using lat/lon
+        this.setRadiusLatLon()
+      }
+    }
   },
   drawingOff() {
     if (this.get('locationType') === 'dms') {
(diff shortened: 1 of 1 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
```diff
@@ -566,12 +574,7 @@ module.exports = Backbone.AssociatedModel.extend({
     const lat = this.get('lat'),
       lon = this.get('lon')
 
-    if (
-      (!Drawing.isDrawing() && this.get('locationType') !== 'latlon') ||
-      !this.isLatLonValid(lat, lon)
-    ) {
-      return
-    }
+    if (!this.isLatLonValid(lat, lon)) return
```
jordanwilking: This allows usng and utmUps to be updated for point radii outside of drawing

jordanwilking:
build now

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
samuelechu: The conditional removed was initially added in this PR: https://github.com/codice/ddf/pull/1875 

(https://github.com/glenhein/ddf/blob/86dceb5ec0515ad4a4389d06e93ab1a5b1fba08c/catalog/ui/catalog-ui-search/src/main/webapp/component/location-old/location-old.js#L310)

There are some concerns related to the point radius in the PR, including https://github.com/codice/ddf/pull/1875#pullrequestreview-34350427. I'm not sure if that conditional is needed or not in preventing a regression somewhere. We might want to follow the test instructions that Glen wrote for point radius testing in the PR above to make sure we didn't regress.

samuelechu (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/location-old/location-old.js
jordanwilking: Good call. I was having trouble tracing this back since the git blame just mentions the move to ddf-ui.

jordanwilking: Looked into this a bit.
1. Looks like the issue I was seeing was related to my changes. The initialization doesn't work without removing these lines, because it prevents the usng and utm fields from being updated.
2. I'm not seeing the issues with drawing that Will ran into, but editing utm is interrupted because this method gets past this point. So, if you start typing coordinates, this stops you and updates your input immediately.

I should be able to just factor out the code I need from this for updating the other coordinate systems and just use it in the init method though, which should avoid any regressions.

jordanwilking:
build now

glenhein (approved):

glenhein:
Hero successful. I was able to search by location and find the expected metacard.

Screen Shot 2021-06-17 at 1 24 50 PM

glenhein:
Screen Shot 2021-06-17 at 1 26 14 PM
Screen Shot 2021-06-17 at 1 26 09 PM
Screen Shot 2021-06-17 at 1 26 05 PM

PR #606, opened 2021-06-17
Updates search type switching to immediately downgrade filterTree if necessary

andrewkfiedler:
- Updates the query model to immediately downgrade filterTrees when switching to basic.  This prevents an issue where if the user happened to not touch the form in anyway after switching, but instead clicked search, it would appear that other terms (visible on advanced) were still around.  Now the type switch downgrades immediately instead of waiting for the user to interact with the component and change something.
 - Also updates the location component to prevent a race condition where the onChangeCallback would be fired when it shouldn't (on destruction).  This would cause us to sometimes lose location information.  Now, it gets pushed onto the stack so that the listener has time to stopListening before we reset things.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.tsx
@@ -348,16 +348,33 @@ const constructFilterFromBasicFilter = ({
   })
 }
 
-const QueryBasic = ({ model }: QueryBasicProps) => {
-  const inputRef = React.useRef<HTMLDivElement>()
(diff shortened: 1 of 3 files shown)


jordanwilking (approved):

jMoneee (approved):

jordanwilking:
build now

jordanwilking:
🔍 Hero Successful 🔍 

#### Steps

1. Cherry-picked https://github.com/codice/ddf-ui/pull/608 for easy ddf setup
2. Created and saved an advanced search with two text fields and a geo field
3. Searched and verified the request was correct
4. Switched to basic, searched, and verified the request
5. Switch to advanced, which was now truncated after being converted to basic, searched, and verified the request
6. Switched back, verified

jordanwilking:
build now

PR #609, opened 2021-06-18
Fix firefox specific issue with resize observer

andrewkfiedler:
- The spec states that the return value should be an array, but firefox will optimize and return the value itself if it's singular.  This update ensures the return is always an array.
 - This could only be seen when trying to access layers on the map in Firefox.  The dropdown could sometimes appear slightly offscreen unless the user triggered a resize themselves.
Screen Shot 2021-06-18 at 1 00 12 PM


jordanwilking (approved):

jlcsmith (approved):

samuelechu (approved):

jMoneee:
Hero Successful
Confirmed the layers dropdown does not go offscreen when opened using the Firefox browser

jlcsmith:
build now

PR #613, opened 2021-06-24
USNG/MGRS Bounding Box Search Fix

jMoneee:
Description

Fixes 2 issues with Bounding Box search drawing and validation of the bounding box. 

The first issue caused bounding box errors to not disappear when a new box was drawn which was fixed by removing some flags that would only allow the validation to occur if there was not an error already present. 

The second error was with the validation of USNG/MGRS bounding boxes in which an `'Upper left coordinate must be located above lower right coordinate'` error would be displayed when it should not. This was caused by string values being passed into the Latitude and longitude validators from the USNG/MGRS validator. The validators would then compare the string values instead of the numbers themselves. I both added handling for strings to the Latitude and longitude validators as well as making sure the USNG/MGRS validator sends a number instead of a string.

Testing

For Error not disappearing issue
1. Build and set up DDF and DDF-UI
2. Create an anyGeo search using a bounding box 
3. Create a Lat/Lon (DD) bounding box with equal North and South coordinates
4. Observe the error
5. Draw new bounding box
6. Verify the error goes away
7. Repeat 3 through 6 for each different coordinate system

For USNG/MGRS validation issue
1. Create a USNG/MGRS bounding box over Australia starting from the bottom left and drawing towards the top right of the country (for example `Upper Left 49L EH 10206 67419 Lower Right 54H WE 08555 30335`)
2. Verify an error is not triggered


andrewkfiedler (approved):

jlcsmith:
build now

samuelechu (approved):

samuelechu:
build now

samuelechu:
Hero successful, verified errors are fixed and that searches with the drawn bounding boxes yielded the expected results.

samuelechu:
build now

PR #615, opened 2021-07-12
Refactors bulk selection for inspector

andrewkfiedler:
- No longer utilizes the bulk edit / action paradigm.
 - Now the user can cycle through using prev / next buttons each selected item.
 - Updates the inspector to use the tabs component from material ui.
 - Updates the various inspector tab views to take in the result directly if possible.
 - Updates the metacard title component to be react (and updates the metacard route to utilize the same component)
 - Fixes a bug in transfer list that applied to both it and the summary view, where the user check on canWrite was only taken into account on remote items.

Screen Shot 2021-07-12 at 10 40 57 AM
Screen Shot 2021-07-12 at 10 41 07 AM


samuelechu (approved):

jordanwilking (approved):

jordanwilking:
🔍 Hero Successful 🔍 

#### Steps

1. Ingest resources and confirmed that there are no more bulk edits
image
2. Verified that I was able to cycle through selected items
3. Updated an item to check the reactified history view. Looks good.
4. Tested the new mui tabs.

PR #618, opened 2021-07-22
Adds case to geo filter generation

jordanwilking:
Treats POINT the same as POINT RADIUS when creating a geo filter.

We don't appear to have point drawing in ddf-ui. This would need to be tested downstream.

We often treat points as point radii with a very small buffer, so they are effectively point radii.


jlcsmith (approved):

andrewkfiedler (approved):

jlcsmith:
Verified downstream

PR #632, opened 2021-10-18
Fix saved search CQL filter value

sgalla:
When saving a search or updating a saved search in Intrigue, the CQL value was not being written correctly due to the filter tree being a string value and not an object. Added logic to attempt to parse the filter tree JSON if it is a string before calling cql.write() on the filter tree. 

Testing:
Save a search and update a saved search. Check the CQL filter value that gets stored for both of these actions and verify the value looks like valid CQL and not like a filter tree.


clockard (approved):

ricklarsen (approved):

shaundmorris (approved):

sgalla:
build now

sgalla:
build now

PR #657, opened 2022-07-12
Removed function that was causing unexpected browser close

wesleyclee:
###
What does this PR do, and why?
- Bugfix change for when opening DDF in a new tab, performing a metacard interaction on returned search result would close the current window.
- Removed unnecessary code that was introducing the bug
###
How should this be tested?
1. Create a simple HTML page to open DDF in a new tab
```
<html>
<body>
<a href="https://isr-cn2a-sys-01.aur.nciteglobal.com.au:8993/" target="_blank">Launch DDF</a>
</body>
</html>
```
 2.  Open the page with Chrome or Firefox, and click on the link.  This will open DDF.
 3.  Navigate to the search tab, and conduct a search that returns at least 1 result.
 4.  Select the result
 5.  Using the kebab menu at the top of the window (next to the heart), perform an action.
 6. The tab no longer closes.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/result-selector.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/result-selector.tsx
@@ -67,9 +67,7 @@ const SelectedResults = ({ selectionInterface }: any) => {
         <Paper>
           <LazyMetacardInteractions
             lazyResults={selectedResultsArray}
-            onClose={() => {
-              close()
-            }}
+            onClose={() => {}}
           />
         </Paper>
       </Popover>


andrewkfiedler (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/result-selector.tsx
```diff
@@ -67,9 +67,6 @@ const SelectedResults = ({ selectionInterface }: any) => {
         <Paper>
           <LazyMetacardInteractions
             lazyResults={selectedResultsArray}
-            onClose={() => {
```
andrewkfiedler, line 70: On close is required as a parameter, so pass in an empty function for now.

andrewkfiedler (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/result-selector/result-selector.tsx
wesleyclee: @andrewkfiedler The change above has been made, can this PR be merged soon?

nnld (commented):
Hero is in progress.

nnld (approved):
Hero succeeded.  All good.

nnld (approved):

nnld (changes requested):
We have found another instance where this issue is manifested.  Potentially, there will be more instances where this method is called.  We will need to fix them all in this PR.

andrewkfiedler:
> We have found another instance where this issue is manifested. Potentially, there will be more instances where this method is called. We will need to fix them all in this PR.

@nnld is there going to be further updates to this PR?

jlcsmith (approved):

wesleyclee:
@andrewkfiedler We are still investigating and testing to find more occurrences of this bug within DDF-UI, and will make the fixes where we think is appropriate and update this PR. After that we will ping you guys for another review

wesleyclee:
@andrewkfiedler We have finished looking into similar cases of the close() function being called unintentionally, and have tested and confirmed that no similar cases exist. This PR is ready to be merged

PR #671, opened 2022-10-07
Fix versioned delete id filter

clockard:
Fix an id filter so it doesn't find deleted discoverable space metacards when deleting deleted space metacards.


clockard:
build now

sgalla (approved):

glenhein (approved):

jlcsmith (approved):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/metacard/MetacardApplication.java
```diff
@@ -1057,7 +1057,8 @@ private void attemptDeleteDeletedMetacard(String id) {
     LOGGER.trace("Attemping to delete metacard [{}]", id);
     Filter tags =
         filterBuilder.attribute(Metacard.TAGS).is().like().text(DeletedMetacard.DELETED_TAG);
```
jlcsmith, line 1059: Not related to this PR, but I see several instances of using like filters on tags in this repo. I have a feeling this wasn't  thought through and probably should be changed to an equals filter.  Thoughts?

PR #672, opened 2022-10-11
Refactor search to allow more complex paging scenarios and start conversion to typescript of the model, adds generic async tasks, adds support for count (page size) to query

andrewkfiedler:
-  Enables usage of our base search for doing advanced paging (next page, last page, better index tracking versus x of y possible - now we can say something like 10-15 if we'd like).
- Updates async tasks to allow generic save / create tasks
- Adds unit tests around paging to prevent future regressions.
- Updates Query model to be typescript so we can have better type checking around it.
- Updates Query to allow custom page sizes (previously we always used the user pref), so even passing a count was simply ignored in the end.


andrewkfiedler:
build now

PR #679, opened 2022-11-16
Fix async task save to use update instead of create

clockard:
Fix the async save to use update instead of create. Using update causes the correct plugin action to be taken in the backend catalog


clockard:
build now

andrewkfiedler (approved):

PR #697, opened 2023-01-18
Use download action provider (if it exists) for UI download

sgalla:
#### What does this PR do, and why?
Update the download functionality to use the download action url if it exists on the result. If the download action does not exist on the result, fallback to the existing download functionality (use the resource-download-url).

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/LazyQueryResult/LazyQueryResult.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/LazyQueryResult/LazyQueryResult.tsx
@@ -333,7 +333,7 @@ export class LazyQueryResult {
   getDownloadUrl(): string {
     const downloadAction = this.plain.actions.find(
       (action) =>
-        action.id === 'catalog.data.metacard.resource.deployed-download'
+        action.id === 'catalog.data.metacard.resource.alternate-download'
     )
 
     return downloadAction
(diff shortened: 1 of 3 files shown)


sgalla:
build now

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/LazyQueryResult/LazyQueryResult.tsx
```diff
@@ -320,8 +320,15 @@ export class LazyQueryResult {
     })
     this.syncWithPlain()
   }
-  isDownloadable(): boolean {
-    return this.plain.metacard.properties['resource-download-url'] !== undefined
+  getDownloadUrl(): string {
+    const downloadAction = this.plain.actions.find(
+      (action) =>
+        action.id === 'catalog.data.metacard.resource.deployed-download'
```
clockard: Realize this isn't really part of this PR but wondering if the action id looked for should be something more generic than "deployed-download" which seems very specific to the downstream use case. Something like `alternate-download` might be a more generic option.

clockard (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/LazyQueryResult/LazyQueryResult.tsx
sgalla: Yeah, that had occurred to me too. I'll bring it up in standup and if everyone agrees I'm happy to change it here and put up a PR in the downstream repo to update the id.

PR #705, opened 2023-01-25
Updates backbone cache handling to be more explicit rather than global, removes dep, moves from $.get to fetch

andrewkfiedler:


ui-frontend/packages/catalog-ui-search/src/main/webapp/js/ApplicationHelpers.ts
```diff
@@ -29,29 +27,3 @@ import 'backbone-associations'
 // This sub object selecting can be see at
 // http://dhruvaray.github.io/backbone-associations/specify-associations.html#sa-getsetop
 Backbone.Associations.setSeparator('>')
-/**
```
andrewkfiedler, line 32: This is unused code.

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/ApplicationSetup.tsx
```diff
@@ -49,30 +43,6 @@ if (process.env.NODE_ENV !== 'production') {
 window['__react-beautiful-dnd-disable-dev-warnings'] = true
 ;(window as any).CESIUM_BASE_URL = './cesium/assets'
 //in here we drop in any top level patches, etc.
-const toJSON = Backbone.Model.prototype.toJSON
```
andrewkfiedler, line 52: More unused code, we no longer use backbone clone in the codebase.

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/Sources.ts
```diff
@@ -56,7 +57,6 @@ const computeTypes = function (sources: any) {
 }
 export default Backbone.Collection.extend({
   url: './internal/catalog/sources',
-  useAjaxSync: true,
```
andrewkfiedler, line 59: Old property that doesn't exist anymore.

sgalla (approved):

PR #706, opened 2023-01-25
Update fetch to account for tests

andrewkfiedler:
- the process check isn't working, so using __ENV__ instead.


andrewkfiedler:
build now

PR #710, opened 2023-01-31
Updates draw button to add color, moves frequency of environment console log

andrewkfiedler:
- Adds primary color to draw button to make it more obvious.
- Moves environment log to be triggered once per app load.


andrewkfiedler:
build now

PR #717, opened 2023-02-16
Allow empty attributes to be hidden in the inspector

jrnorth:
https://user-images.githubusercontent.com/4495447/219449088-e2f92a7d-1334-4caf-bddc-55d2d2528f32.mp4

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
@@ -432,17 +432,28 @@ export const Editor = ({
 const AttributeComponent = ({
   lazyResult,
   attr,
+  showEmpty,
   summaryShown = [],
   filter = '',
   forceRender,
 }: {
   attr: string
   lazyResult: LazyQueryResult
+  showEmpty: boolean
   summaryShown?: string[]
   filter?: string
   forceRender: boolean
 }) => {
   let value = lazyResult.plain.metacard.properties[attr]
+  if (!showEmpty) {
+    if (!value) {
+      return null
+    } else if (typeof value === 'string' && !value.trim()) {
+      return null
+    } else if (Array.isArray(value) && value.length === 0) {
+      return null
+    }
+  }
   if (value === undefined || value === null) {
     value = []
   }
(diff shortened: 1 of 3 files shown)


alexabird (approved):

jlcsmith (approved):

andrewkfiedler:
We reuse the transfer list component in the results listing (both the table / list have their own pref), so we'll want to make sure we don't show that control when it's opened from those components.  
Screen Shot 2023-02-16 at 10 07 59 AM

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
```diff
@@ -432,17 +432,28 @@ export const Editor = ({
...
   attr: string
   lazyResult: LazyQueryResult
+  showEmpty: boolean
   summaryShown?: string[]
   filter?: string
   forceRender: boolean
 }) => {
   let value = lazyResult.plain.metacard.properties[attr]
+  if (!showEmpty) {
+    if (!value) {
```
andrewkfiedler: We might want to be explicit here (`value === undefined || value === null`), I can't remember if value can ever be a number but if so we'll miss 0.

andrewkfiedler (changes requested):
Left a few comments, but looks good to me.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/summary.tsx
jrnorth: Good point, I'll update that.

jrnorth:
@andrewkfiedler That's what this check is for: https://github.com/codice/ddf-ui/blob/ab15f13584b778e65a1f25791ab96344bd3e40dc/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/transfer-list.tsx#L825
The summary component is the only one that sets the prop. I updated the demo video to demonstrate.

andrewkfiedler:
> @andrewkfiedler That's what this check is for:
> 
> https://github.com/codice/ddf-ui/blob/ab15f13584b778e65a1f25791ab96344bd3e40dc/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/tabs/metacard/transfer-list.tsx#L825
> 
> The summary component is the only one that sets the prop. I updated the demo video to demonstrate.

Gotcha, the boolean or undefined there was confusing me.  I was like. . .how, then I realized you're checking if it's undefined.

andrewkfiedler (approved):

alexabird:
✅ Hero passed - verified this toggle only shows in Manage Attributes when clicked from inspector view

PR #726, opened 2023-03-01
Update callback for golden layout state change to handle detachment

andrewkfiedler:
- Previously the user could cause this callback to trigger, then in some cases leave the page or otherwise make the component disappear.  In that scenario, the config would be reported as empty when it wasn't.
 - Now we still debounce the update, but call for the config immediately on state change, avoiding the possibility that we getthe wrong config from golden layout.
 - Also fixes a minor issue where starting from a blank layout, then adding a visual, didn't cause a save to happen.  Subsequent changes did fire correctly, but this was solvable by moving the initialization of lastConfig to the init event instead of the first change event.

Testing procedure:
1.  Make a layout with any number of visuals, one will do.
2.  Quickly go back and forth between a page with the visuals and one without.
3.  Verify the visuals never end up being saved as empty.

(on master, you can verify this is reproducible as a bug in the same way)

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
@@ -345,39 +345,34 @@ export function getInstanceConfig({ goldenLayout }: { goldenLayout: any }) {
 function handleGoldenLayoutStateChange({
   options,
   goldenLayout,
+  currentConfig,
   lastConfig,
 }: {
   goldenLayout: any
+  currentConfig: any
   options: GoldenLayoutViewProps
   lastConfig: React.MutableRefObject<any>
 }) {
-  if (lastConfig.current === null) {
-    // this triggers on init of golden layout
-    lastConfig.current = getInstanceConfig({ goldenLayout })
-    return
-  }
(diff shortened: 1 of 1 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
```diff
@@ -535,19 +530,43 @@ export const GoldenLayoutViewReact = (options: GoldenLayoutViewProps) => {
...
             options,
+            currentConfig,
             goldenLayout,
             lastConfig,
           })
-        }, 200)
+        },
+        200
       )
+      /**
```
andrewkfiedler, line 545: I debated trying to solve this as well, but it seemed complicated.  I tried hooking into item destroyed, but that triggers on leaving the page and causes more issues than it solves.  And unfortunately, golden layout (version 1 at least), is out of active development.

Thankfully, this is quite the edge case.  We're only missing the saving of a completely empty layout, and only in some cases.

sgalla (approved):

sgalla:
✅ Hero Successful! I was able to reproduce the bug on master and verify that it is fixed with these changes.

sgalla:
build now

alexabird (approved):

PR #729, opened 2023-03-08
Get sort from user preferences for table-export

alexabird:
Untitled copy
TESTING:
Upload multiple products, and run a * search. Click export > select CSV. Confirm that the export file contains the expected results and they they are in the same order as in the Results pane. Try a few different sorting configurations.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/table-export/table-export.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/table-export/table-export.tsx
@@ -82,8 +82,11 @@ function getColumnOrder(): string[] {
 function getHiddenFields(): string[] {
   return user.get('user').get('preferences').get('columnHide')
 }
-function getSorts() {
-  return user.get('user').get('preferences').get('resultSort')
+function getSorts(selectionInterface: any) {
+  return (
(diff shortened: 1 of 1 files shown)


jrnorth (approved):

alexabird:
build now

pvargas (approved):

swirepe (approved):

sgalla:
With your changes, will the exported results respect the query sort order if there are not any result sorts set?

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/table-export/table-export.tsx
```diff
@@ -81,6 +82,9 @@ function getColumnOrder(): string[] {
 function getHiddenFields(): string[] {
   return user.get('user').get('preferences').get('columnHide')
 }
+function getSorts() {
+  return user.get('user').get('preferences').get('resultSort')
```
sgalla: With this change, will the exported results be sorted according to the query sorts if the user has not set any result sorts?

alexabird: It does now 🙃 thanks for catching that!

PR #730, opened 2023-03-13
Add custom enter key submit event, listeners for triggering searches

andrewkfiedler:
- Adds custom enter key submit event
 - Adds some utilities around it, for common handlers, props, hooks.
 - Updates the search experience so that pressing enter in certain fields, like text, date, number executes the search.
 - Each component can customize when it fires the custom event, so dropdowns can avoid triggering unless they are closed, etc.
 - Fixes an issue where we weren't executing on enter in the right area filter (the dropdown for extra filters next to layout).  There was some old code here that wasn't working, so I replaced it.  Now pressing enter here when an execution would happen will immediately close and execute the search with the new additional filters.
 - Updates the boolean search to not autofocus by default, since that's not always the desired behavior.
 - Updates the text input filter to more gracefully handle coming in with a value originally from the boolean filter.  This is one way for now, since going from text to boolean is a bit more complicated (we have to ask the server to validate it, etc.).

Things to test:
- Verify the enter key functionality works on the various text inputs as expected (not triggering when dropdowns are open).  Test contains, =, near, boolean, etc.  Also verify it works on enums, like datatype.
- Verify that the enter key does **not** do anything for location inputs. 
- Verify that the enter key works on date inputs.

https://user-images.githubusercontent.com/11984853/224836790-1d825b93-eabb-4260-a581-fac8f85dd3ef.mp4


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/boolean-search-bar/boolean-search-bar.tsx
```diff
@@ -158,16 +159,19 @@ const BooleanSearchBar = ({
     if (option.length === 0) return ''
     return option.token
   }
-  const handleSubmit = () => {
-    if (!value.error && props.onSubmit) {
-      props.onSubmit()
+  const handleSubmit = (e: React.SyntheticEvent) => {
+    if (!value.error && (!isOpen || currentOptions.length === 0)) {
+      console.log('submit the query')
```
sgalla: extra log?

sgalla (approved):

alexabird (approved):

PR #738, opened 2023-03-30
Upgrade to webpack 5, node 16

andrewkfiedler:
- Upgrade to latest webpack (through ace dependency)
 - Upgrade tailwind from v1 to v3, note I added a custom preflight since there are issues with buttons otherwise
 - Upgrade polished
 - Upgraded typescript / prettier
 - This also updates our tests to be run on tsx rather than compiled jsx


andrewkfiedler:
build now

andrewkfiedler:
build now

ui-frontend/packages/catalog-ui-search/package.json
```diff
@@ -30,7 +30,7 @@
     "url": "https://github.com/codice/ddf.git"
   },
   "engines": {
-    "node": ">=0.10.5"
+    "node": ">=16.16.0"
```
figliold: ❓ I'm using node version 17.3.1.  For the node commands in `package.json` I've had to add the `--openssl-legacy-provider` argument.  Is it safe to use `>=16.16.0` here without that argument?

andrewkfiedler: No clue, might be worth locking it down further if we have dependencies that can't handle higher versions yet.  I'll go ahead and do that.

jrnorth (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/index.html
```diff
@@ -24,7 +24,7 @@
     <meta name="description" content="">
     <meta name="author" content="">
 
-    <link id="favicon" rel="icon" type="image/png" href="../../data:image/png;base64,">
+    <!-- <link id="favicon" rel="icon" type="image/png" href="../../data:image/png;base64,"> -->
```
sgalla: Can this be removed or uncommented?

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/ApplicationSetup.tsx
```diff
@@ -28,12 +28,12 @@ import '../styles/plotly.css'
 import Backbone from 'backbone'
 import properties from './properties'
 import './extensions/application.patches'
-import '@connexta/icons/icons/codice.font'
+// import '@connexta/icons/icons/codice.font'
```
sgalla: Extra comment?

ui-frontend/packages/catalog-ui-search/yarn-error.log
sgalla, line 1: Should this file be git ignored?

sgalla (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/index.html
andrewkfiedler: I'll take a look, it doesn't look like we're doing anything with favicons at the moment, which is a bit odd.

andrewkfiedler: Yeah, I've looked and can't find any reference that uses this in the code.  And it doesn't look like we have a favicon setup on the existing systems that use it.  So maybe that's something we need to fix, but for now I'll remove this.

PR #740, opened 2023-04-04
Fix coordinate copying on map

jrnorth:
The clipboard library we were using worked by inserting a textarea with the desired text, then selecting it and calling execCommand('copy'). Some time ago, the copy-coordinates component was updated to use a modal, which prevented the fake textarea element from being focused normally. There are ways to work around this, but we elected to switch to the Clipboard API instead, to ensure more consistent behavior across browsers.

### Testing Instructions
Copy all the supported coordinate formats on both the 2D and 3D maps, in at least Chrome and Firefox.


alexabird (approved):

pvargas (approved):

alexabird:
build now

PR #743, opened 2023-04-07
use new preferences

glenhein:
This requires a new DDF: https://github.com/codice/ddf/pull/6732


clockard (approved):

clockard:
If you switch to a single field holding the json blob as mentioned in my comment in the ddf pr I don't think you would need to worry about the metacard injections and this could stand on its own.

clockard:
build now

clockard:
build now

rececoffin (approved):

PR #751, opened 2023-04-24
Create an audit log entry for searches that includes all selected sources

glenhein:
Since the UI breaks searches into two backend calls (one for harvested and one for non-harvested), there wasn't a single audit log entry that included all of the selected sources. This PR calls the audit log endpoint to create a new audit log entry with all of the selected sources.

Example:

```
[INFO ] 2023-04-24T05:58:58,663 | tp1767707945-801 | securityLogger  |  trace-id 8966a4f167044cb59329faac2c3c1b23, Subject: admin,  Client IP: 127.0.0.1, Port: 60302, SEARCH_SUBMITTED a query: [((anyText ILIKE '*'))] sources: [EARTHQUAKES,GEOSERVER,BAR,CAT,DOG,FOO,HARVESTED SOURCE,ISRI,WITH SPACE]
```

### DIFF
--- a/ui-backend/audit/audit-application/src/main/java/org/codice/ddf/catalog/audit/application/SimpleAuditHandler.java
+++ b/ui-backend/audit/audit-application/src/main/java/org/codice/ddf/catalog/audit/application/SimpleAuditHandler.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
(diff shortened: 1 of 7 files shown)


ui-backend/audit/audit-application/src/main/java/org/codice/ddf/catalog/audit/application/SimpleAuditHandler.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  public void handle(@NotNull Context context) {
+    AuditRequestBasic requestBasic = context.bodyAsClass(AuditRequestBasic.class);
+
+    try {
+      auditService.log(requestBasic.getAction(), requestBasic.getComponent());
+      context.status(HttpStatus.SC_OK);
+    } catch (AuditException e) {
+      LOGGER.error("Unable to log the user's action. Error message: {}", e.getLocalizedMessage());
```
jlcsmith: ❓ any particular reason you are logging the message and not the exception itself?

jlcsmith (approved):

sgalla (approved):

ui-backend/audit/audit-application/src/main/java/org/codice/ddf/catalog/audit/application/SimpleAuditHandler.java
glenhein: fixed. The original code in AuditHandler was doing the same. I updated my code and the original.

PR #756, opened 2023-04-27
Improve location input error handling

jrnorth:
This PR enhances the location inputs, which now accept a prop (`errorListener`) that can be used to notify interested parties of validation errors. It also eliminates the use of `onBlur` for validation, so users can get immediate feedback as they type. Finally, it makes several improvements to location validation to handle more cases.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/utils/validation/validation.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/utils/validation/validation.tsx
@@ -267,7 +267,10 @@ function getGeometryErrors(filter: any): Set<string> {
 }
 function validateLinePolygon(mode: string, currentValue: string) {
   if (currentValue === undefined) {
-    return initialErrorState
+    return {
+      error: true,
+      message: `${mode === 'line' ? 'Line' : 'Polygon'} cannot be empty`,
+    }
   }
   try {
     const parsedCoords = JSON.parse(currentValue)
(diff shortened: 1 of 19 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/utils/validation/validation.tsx
```diff
@@ -521,12 +534,30 @@ function validateUtmUps(key: string, value: any) {
   return error
 }
 function validateRadiusLineBuffer(key: string, value: any) {
+  const parsed = Number(value.value)
+  const buffer = Number.isNaN(parsed) ? 0 : parsed
+  const bufferMeters = DistanceUtils.getDistanceInMeters(buffer, value.units)
+  if (key === 'lineWidth' && bufferMeters < 75) {
```
jlcsmith: might be worth adding a comment for posterity on why 75 was chosen...  "75 meters determined to be a reasonable min for linestring searches based on data source capabilities"  (or similar)

jlcsmith (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/utils/validation/validation.tsx
jrnorth: Done.

pvargas (approved):

alexabird (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/utils/validation/validation.tsx
```diff
@@ -116,9 +121,17 @@ export const ErrorComponent = (props: any) => {
   ) : null
 }
 export function validateListOfPoints(coordinates: any[], mode: string) {
-  let message = ''
+  let message = null
   const isLine = mode.includes('line')
   const numPoints = isLine ? 2 : 4
```
pvargas, line 126: I know you didn't introduce this, but I think this should be changed from 4 to 3, since a triangle is a valid polygon.

pvargas:
Other than the mentioned above, hero was successful in downstream project.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/utils/validation/validation.tsx
jrnorth: The 4 is because a polygon requires the first and last points to be the same for it to be closed. The number of unique points is the length of the coordinates array minus one.

PR #759, opened 2023-05-10
Fix some WMTS issues on the 2D map

jrnorth:
- ensure the `request=GetCapabilities` query parameter is set on the URL when fetching the capabilities
- the layer and matrixSet options are now always set properly on the WMTS source
- when the proxy is enabled, it is now used for all WMTS requests, not just GetCapabilities


jrnorth:
@regan-prendergast
@dantolini

jlcsmith (approved):

alexabird (approved):

jrnorth:
tested by @regan-prendergast

PR #761, opened 2023-05-12
Change color of selected map pin

alexabird:
As part of this, we also want to scroll to the selected pin in the results list. I'm pretty sure https://www.npmjs.com/package/react-window-components needs to be updated, but I don't have access to the repository.
@andrewkfiedler 

image


pvargas (approved):

jlcsmith:
🎉 Hero Successful 🎉 
- Verified pin highlight on selection in 2D/3D map
- Verified map pins for various metacard geometries
- verified light/dark themes
- verified result list selection based on selected pin
image

PR #763, opened 2023-05-18
Update padding in various areas to improve UX

andrewkfiedler:
- Update result selector to have less padding (is above the golden layout visualizations)
 - Update basic query view to have less padding
 - Update a few usages of Grid to instead use tailwind + normal divs
 - Update how we construct the Theme so that we have type checking.  In addition, adds some type declaration for the addition we made of the color grey previously.  Also updates the default size for buttons to be small.
 - Update the index.html so dark theme is assumed until preferences load.  This is less jarring for dark mode users.  And for light mode users, seeing dark mode for a moment is fine.
 - Update tailwind config to include html files as part of it's content determination


jlcsmith (approved):

jlcsmith:
Hero looked good - verified padding changes per the description.

PR #770, opened 2023-06-07
Update supported browsers

andrewkfiedler:
- Updates our supported browsers to be inline with https://mui.com/material-ui/getting-started/supported-platforms/#browser
- Change from feature detection to looking at user agent string to determine browser + version.  In the future this may get deprecated, so we might have to migrate it then.  But I couldn't find a feature overlapping those browser versions to test.
- Adds a "continue" button to the unsupported screen in case folks want to use the application anyways.  It sets a session flag, so each time the browser is reopened they'll have to confirm again.  

Screenshot 2023-06-06 at 6 27 09 PM
Screenshot 2023-06-06 at 6 27 38 PM

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/check-browser.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/check-browser.tsx
@@ -58,6 +58,12 @@ export const handleUnsupportedBrowser = () => {
     })
 }
 
+export const handleSupportedBrowser = () => {
+  document.querySelector('#incompatible-browser-bg')?.remove()
+}
+
 if (!isSupportedBrowser()) {
   handleUnsupportedBrowser()
+} else {
+  handleSupportedBrowser()
 }
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/check-browser.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/check-browser.tsx
@@ -44,6 +44,11 @@ export const isSupportedBrowser = () => {
         return false
     }
   } else {
+    /**
+     *  Assumption is that this would happen only in a new browser version where the user agent deprecation
+     *  has gone through (thus it should be a relatively new version of chrome or whatever browser),
+     *  so we let it through as supported.
+     */
     return true
   }
 }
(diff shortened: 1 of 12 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/check-browser.tsx
```diff
@@ -0,0 +1,69 @@
...
+      case 'edge':
+      case 'edge-chromium':
+        // https://officecdnmac.microsoft.com/pr/03adf619-38c6-4249-95ff-4a01c0ffc962/MacAutoupdate/MicrosoftEdge-90.0.818.39.pkg
+        // change version on end to get to old versions
+        return version >= 91
+      default:
+        return false
+    }
+  } else {
+    return true
```
andrewkfiedler, line 52: I could be convinced that we should default to false if the user agent detection fails.  My thought here is that it would only happen when the deprecation goes through, so it should be a relatively new version of chrome or whatever browser, thus we let it through.

jrnorth: I agree with your reasoning for letting it through. We should add that explanation as a comment here.

ui-frontend/packages/catalog-ui-search/src/main/webapp/check-browser.tsx
```diff
@@ -0,0 +1,69 @@
...
+        // https://chromium.cypress.io/mac/ for those devs who wish to test old chromium versions
+        return version >= 90
+      case 'firefox':
+        // https://ftp.mozilla.org/pub/firefox/releases/
+        return version >= 78
+      case 'edge':
+      case 'edge-chromium':
+        // https://officecdnmac.microsoft.com/pr/03adf619-38c6-4249-95ff-4a01c0ffc962/MacAutoupdate/MicrosoftEdge-90.0.818.39.pkg
+        // change version on end to get to old versions
+        return version >= 91
```
jrnorth, line 42: Just curious how we arrived at these version numbers.

jrnorth (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/check-browser.tsx
andrewkfiedler: They're taken straight from https://mui.com/material-ui/getting-started/supported-platforms/#browser

alexabird (approved):

alexabird:
build now

PR #774, opened 2023-06-16
Fixes surefire and jacoco plugin configurations

jrnorth:
Previously, the JaCoCo plugin was unable to check our code coverage because we configured it and the surefire plugin incorrectly. I don't know how long our coverage has not been checked.

The fix required several changes:
1. Add the prepare-agent execution to the jacoco plugin so it can set the argLine property
2. Add the @{argLine} late replacement to the surefire plugin so it uses the argLine set by the jacoco plugin
3. Don't set forkCount to 0 in the surefire plugin, as that prevents the tests from running with the javaagent set (see
https://www.jacoco.org/jacoco/trunk/doc/maven.html)

Since JaCoCo is actually checking our code coverage now, I had to reduce the limits in several modules for the build to pass.

Several improvements unrelated to the fix were added:
1. Upgrade the surefire and jacoco plugins to their latest versions
2. Configure the jacoco plugin to produce coverage reports
3. Configure the default limits to 0.75
4. Configure the limits to use codice-maven's LenientLimit


jrnorth:
build now

alexabird (approved):

andrewkfiedler (approved):

PR #778, opened 2023-06-20
New extension points for customizing inspector title and map badges

sgalla:
Adds new extension points to customize the metacard title in the inspector visual and to add custom badges to map markers. The map badges are used for both pin and cluster markers in the 2D and 3D map visuals.


sgalla:
build now

PR #779, opened 2023-06-21
Update rpc to wait for connection before being set as available

andrewkfiedler:
- We were checking if sockets were possible, but not if the connection went through without issue.  Now we wait for the connection to open before setting the rpc as available.

Testing steps: 
1.  Use firefox, and go to about:config, turning max web socket connections to 0 (remember to undo this later!)  
Screenshot 2023-06-21 at 8 27 47 AM

2.  Go to an individual metacard's page.  Refresh and verify it loads correctly, finding the metacard.


sgalla (approved):

derekwilhelm (approved):

sgalla:
✅ Hero Successful!

PR #800, opened 2023-08-08
Add ability to allow popping visuals out into their own window

andrewkfiedler:
This is the breakdown by commit, which might be easier to review.

- Added popout/cross-window communication to golden layout usage
  - Implemented router and local storage patches for communication and refreshes between windows.

- Remove unused event types

- Add preference syncing between subwindows and main window
  - Implemented preference synchronization between golden-layout subwindows and main window.

- Add default for resultCount to properties
  - Set a default value for resultCount in properties to prevent issues during page loading.

- Fix useSelectionOfLazyResult hook for changing lazyResult
  - Adapted hook to handle changes in lazyResult, especially in subwindows.

- Remove unnecessary conditions around state change callbacks
  - Simplified event emission conditions with state change comparisons.

- Update removeEphemeralState to remove popout dimensions
  - Removed unreliable popout dimensions tracking to avoid unnecessary resizing.

- Fix updates not syncing between windows due to shared objects
  - Deep cloning implemented to prevent updates on shared objects across windows.
  - Updated cloning method for better performance.

- Update geometry for correct updates when lazyResult changes
  - Fixed updates not reflecting due to missing dependencies in useEffects.

- Start looking at stack support in subwindows
  - Introduced disabling to prevent subwindows of subwindows.
  - Refined handling of layout changes in subwindows.

- Add popout tab button and fix histogram update issue
  - Added popout tab button in main windows.
  - Addressed histogram update issue during initialization.

- Add support for subwindows communicating layout changes
  - Enabled subwindows to notify the main window about layout changes without sending a config.

- Add listening for subwindow close events
  - Improved handling of subwindow close events by latching onto windowClose and forwarding the event source.

- Add communication of subwindow navigation events to main window
  - Subwindow navigation blocked, and events sent to main window for proper handling.
  - Updated components to use React Router directly and removed old backbone router compatibility.

- Add handling of subwindows without parents
  - Popup dialog implemented, requesting users to close parentless subwindows.

- Add handling for popups being blocked
  - Addressed scenarios where popups are blocked in a visualization layout with popouts.
  - Users prompted to enable popups and retry or proceed without popups.

- Refactor highlights storage and communication between subwindows
  - Highlights stored in LazyQueryResults and shared between subwindows.
  - Moved transformations to separate functions.
  - Improved typesafety in QueryResponse.
  - Handled highlights effectively between subwindows.

- Add handling for showingResultsForFields and didYouMeanFields
  - Ensured preservation of these fields when navigating between the main window and subwindows.

- Update anchor handling for blob URLs in subwindows
  - Updated handling of blob URLs to prevent forwarding, as they're typically used for downloads.


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
```diff
@@ -557,23 +653,388 @@ export const GoldenLayoutViewReact = (options: GoldenLayoutViewProps) => {
...
+          ) !==
+          JSON.stringify(
+            Object.values(eventData.lazyResults.results).map((lazyResult) => {
+              return {
+                plain: lazyResult.plain,
+                isSelected: lazyResult.isSelected,
+              }
+            })
+          )
+        ) {
```
jrnorth: ✏️ Recommend something like this instead, to improve readability:
```typescript
const results = Object.values(lazyResults.results).map((lazyResult) => {
  return {
    plain: lazyResult.plain,
    isSelected: lazyResult.isSelected,
  }
})
const callbackResults = Object.values(eventData.lazyResults.results).map((lazyResult) => {
  return {
    plain: lazyResult.plain,
    isSelected: lazyResult.isSelected,
  }
})
if (!_.isEqual(results, callbackResults)) {
```

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
```diff
@@ -557,25 +861,502 @@ export const GoldenLayoutViewReact = (options: GoldenLayoutViewProps) => {
...
+
+  useCrossWindowGoldenLayoutCommunication({
+    goldenLayout,
+    isInitialized: !error && finished,
+    options,
+  })
+
   return (
     <div data-element="golden-layout" className="is-minimised h-full w-full">
+      {popupHandlingState === 'blocked' ? (
```
jrnorth, line 1352: ✏️ `{popupHandlingState === 'blocked' && (`

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
```diff
@@ -135,6 +158,19 @@ const GoldenLayoutComponentHeader = ({
             />
           ) : null}
         </Grid>
+        <Grid item>
+          {!tab.contentItem.layoutManager.isSubWindow &&
+          tab.closeElement[0].style.display !== 'none' ? (
```
jrnorth, line 164: ✏️ Could use another `&&` instead of a ternary.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
```diff
@@ -454,6 +671,19 @@ const GoldenLayoutToolbar = ({ stack }: { stack: any }) => {
                 <AllOutIcon />
               </Button>
             </Grid>
+            {stack.layoutManager.isSubWindow ? null : (
```
jrnorth, line 675: ✏️ `{!stack.layoutManager.isSubWindow && (`

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
```diff
@@ -135,6 +158,19 @@ const GoldenLayoutComponentHeader = ({
...
           ) : null}
         </Grid>
+        <Grid item>
+          {!tab.contentItem.layoutManager.isSubWindow &&
+          tab.closeElement[0].style.display !== 'none' ? (
+            <Button
+              data-id="popout-tab-button"
+              onClick={() => {
+                tab.contentItem.popout()
+              }}
```
jrnorth, line 169: ✏️ `onClick={tab.contentItem.popout}`

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
```diff
@@ -151,11 +187,165 @@ const GoldenLayoutComponentHeader = ({
...
+
+  if (missingWindow) {
+    return (
+      <Dialog open={true} className=" z-[1000000]">
+        <DialogTitle>Could not find parent visualization</DialogTitle>
+        <DialogContent>Please close the window.</DialogContent>
+        <DialogActions>
+          <Button
+            variant="contained"
+            onClick={() => {
```
jrnorth, line 286: ✏️ `onClick={window.close}`

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
```diff
@@ -454,6 +671,19 @@ const GoldenLayoutToolbar = ({ stack }: { stack: any }) => {
                 <AllOutIcon />
               </Button>
             </Grid>
+            {stack.layoutManager.isSubWindow ? null : (
+              <Grid item>
+                <Button
+                  data-id="popout-layout-button"
+                  onClick={() => {
+                    stack.popout()
+                  }}
```
jrnorth, line 681: ✏️ `onClick={stack.popout}`

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/cach-bust-url.tsx
jrnorth, line 1: Filename should be `cache-bust-url.tsx`

jrnorth (approved):
Great work, all the commits and code are very well-commented.

alexabird (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
andrewkfiedler: I've never really liked doing the && style of conditional rendering.  I didn't really have a good reason why, other than it wasn't obvious to me.  So I googled to see if there was a more comprehensive discussion on it.

I think either are okay, but it looks like folks say there is a possibility of messing up with the && if we aren't careful around falsey things (these cases in the code are explicit though, so it doesn't apply).

https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

I think for me it's more that I like how clear it is to see it's a conditional render, and later on if we wanted we could return something else instead of null.  That and I'd be worried that as we add on to a && instance later on, we might mess up with the falsey stuff.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
andrewkfiedler: So one of the reasons I don't do this, is a lot of old libraries tend to depend on `this` binding, and this would muck with that.  I double checked to see if that was the case here, and it indeed does depend on on this, throwing an error if we try to set it as a direct call.

So in general I tend to not set onClick to function calls directly (unless it's our custom code, but even then, I'd rather not think, despite modern js not relying as much on implicit this binding), just so I don't have to think about it.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
jrnorth: Good article. I might start writing it that way, too.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/golden-layout.view.tsx
jrnorth: Good point. I didn't know that.

alexabird:
build now

alexabird:
build now

PR #803, opened 2023-08-11
Refactor color tool styling

sgalla:
The use of withStyles was causing some issues in downstream projects after the MUI upgrade. This PR refactors the ColorTool component so it is still styled as intended, but without the use of withStyles.


sgalla:
build now

pvargas (approved):

jlcsmith (approved):

PR #809, opened 2023-08-25
Add New Attribute to Query Metacard 

pvargas:
This PR adds the `searchAreaIds` attribute to the query metacard.


alexabird:
build now

alexabird:
build now

alexabird (approved):

jrnorth:
This will work, but we may want to consider making this change downstream using attribute injection instead, since nothing in DDF or DDF-UI will be using this (and we won't have to re-release DDF-UI). Anyone have thoughts on that?

PR #811, opened 2023-09-06
Location checkbox - Label and Placeholder updated

shiv-leidos:
Label for Location checkbox is now looked up from location alias in configuration file "org.codice.ddf.catalog.ui.attribute.aliases.config"

Old UI
image

Update UI
image

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.tsx
@@ -474,7 +474,7 @@ const QueryBasic = ({ model, errorListener }: QueryBasicProps) => {
                 }}
               />
             }
-            label="Location"
+            label={TypedMetacardDefs.getAlias({ attr: 'location' })}
           />
           {basicFilter.anyGeo[0] ? (
             <Grid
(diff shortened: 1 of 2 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.tsx
```diff
@@ -474,7 +474,7 @@ const QueryBasic = ({ model, errorListener }: QueryBasicProps) => {
                 }}
               />
             }
-            label="Location"
+            label={TypedMetacardDefs.getAlias({ attr: 'location' })}
```
jlcsmith, line 477: What happens if there is no alias defined for location?

shiv-leidos: In absence of alias, "location" is used as label.

jlcsmith (approved):

pvargas (approved):

maiiiiiiii:
hello,
I did a hero on this and I had a comment: 
1. very minuscule update, but can we get the defaulted "location" to capitalized "Location" to be synonymous with the rest of the labels? Unless you don't foresee a scenario where the label would ever be defaulted. E.g. "Time" "Types"
Screenshot 2023-09-08 at 12 51 43 PM

otherwise, everything checks out :)

shiv-leidos:
Hi Mai,

This is as per exiting practice of ddf-ui. If attribute's alias is not configured, alias is displayed. For Example:

image

image

PR #812, opened 2023-09-13
 Remove old minimize behavior, add new minimize behavior 

andrewkfiedler:
- Removes the old minimize behavior
 - Adds a new minimize behavior for stacks / tabs that is similar to browser windows, but collapses things into a singular stack (working within the limitations of golden layout).
 - There are some workarounds, documented in the code, to addChild not properly removing the existing content as golden layout documentation suggests.  Instead we have to remove things, and add them.  Lots of imperative code that depends on mutating state, so it's a bit complex, but broken down into various functions to make things hopefully easier to digest.
 - Added some tooltips to make some of the buttons clearer, and updated the icons for fullscreen / exit fullscreen to be more standard.


alexabird (approved):

pvargas (approved):

maiiiiiiii:
hero in progress. i can't update the label :'(

maiiiiiiii:
hero was successful!

alexabird:
pulled master and ran full build

PR #823, opened 2023-10-11
Enable double click to edit a geometry

alexabird:
* Double click on geo to go into edit mode
* press enter to apply changes
* press escape to cancel
* when hovering over a filter geo show the "grab" cursor
* when overing over a geo puled in from an area, show the "not-allowed" cursor

To make it so the drawn Locations (the ones you double click on) could reference the original filter Location (the ones that actually get edited), created a locationID and applied this to all location generated from the original filter location. Only set this locationID on filter Locations (not locations populated from areas) and used the existence of locationID to determine the hover cursor (editablilty).

https://github.com/codice/ddf-ui/assets/14789712/4d353e7f-7d69-4706-92c9-7038f2a717de

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
@@ -103,7 +103,7 @@ export const CesiumDrawings = ({
         cancelDrawing()
       }
     },
-    [drawingModel]
+    [drawingModel, drawingShape, drawingLocation]
   )
 
   useEffect(() => {
@@ -116,6 +116,7 @@ export const CesiumDrawings = ({
     } else {
       window.removeEventListener('keydown', handleKeydown)
     }
+    return () => window.removeEventListener('keydown', handleKeydown)
   }, [drawingModel])
 
   const cancelDrawing = () => {
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
@@ -93,12 +93,28 @@ export const CesiumDrawings = ({
   const [isDrawing, setIsDrawing] = useState(false)
   const [drawingShape, setDrawingShape] = useState<Shape>(DEFAULT_SHAPE)
 
+  const handleKeydown = React.useCallback(
+    (e: any) => {
+      if (e.key === 'Enter') {
+        e.preventDefault()
+        finishDrawing()
+      }
+      if (e.key === 'Escape') {
+        cancelDrawing()
+      }
+    },
+    [drawingModel]
+  )
+
   useEffect(() => {
     setIsDrawing(!!drawingModel)
     if (drawingModel) {
+      window.addEventListener('keydown', handleKeydown)
       setDrawingShape(
         getShapeFromDrawMode(getDrawModeFromModel({ model: drawingModel }))
       )
+    } else {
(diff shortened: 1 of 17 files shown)


maiiiiiiii:
hero successful!

andrewkfiedler:
Cool!

andrewkfiedler (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
```diff
@@ -93,12 +93,28 @@ export const CesiumDrawings = ({
   const [isDrawing, setIsDrawing] = useState(false)
   const [drawingShape, setDrawingShape] = useState<Shape>(DEFAULT_SHAPE)
 
+  const handleKeydown = React.useCallback(
```
jrnorth, line 96: ❓What's the value in using `useCallback()` here?

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
```diff
@@ -93,12 +93,28 @@ export const CesiumDrawings = ({
...
+
   useEffect(() => {
     setIsDrawing(!!drawingModel)
     if (drawingModel) {
+      window.addEventListener('keydown', handleKeydown)
       setDrawingShape(
         getShapeFromDrawMode(getDrawModeFromModel({ model: drawingModel }))
       )
+    } else {
+      window.removeEventListener('keydown', handleKeydown)
```
jrnorth, line 117: This effect should also return a cleanup function that removes the event listener.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/location/location.tsx
```diff
@@ -139,6 +139,11 @@ const LocationInput = ({ onChange, value, errorListener }: any) => {
     locationModel.set('color', color)
     ;(wreqr as any).vent.trigger('search:drawend', [locationModel])
   }
+  listenTo((wreqr as any).vent, 'location:doubleClick', (locationId: any) => {
```
jrnorth: I think we'll want to use an effect for this, similar to the onChangeCallback, so we can remove the listener when this component is removed from the DOM.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/drawing-and-display.tsx
```diff
@@ -305,13 +305,29 @@ export const OpenlayersDrawings = ({
...
   useEffect(() => {
     setIsDrawing(!!drawingModel)
     if (drawingModel) {
+      window.addEventListener('keydown', handleKeydown)
       setDrawingShape(
         getShapeFromDrawMode(getDrawModeFromModel({ model: drawingModel }))
       )
       setDrawingGeometry(getDrawingGeometryFromModel(drawingModel))
+    } else {
+      window.removeEventListener('keydown', handleKeydown)
```
jrnorth, line 330: Same here, I think this effect needs a cleanup function to remove the listener.

jrnorth (commented):
nice solution 👍🏻

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
alexabird: Just defining the function outside of the useEffect doesn't work and I thought useCallack was more readable than putting the function in the useEffect (can tell which state vars it dependent on)

jrnorth (approved):

PR #835, opened 2023-10-24
added ability to create custom datatype query configuration using the…

kcwire:
… extraJson config option


alexabird (approved):

alexabird:
I might have something misconfigured for the hero.
With `metadata-content-type` configured as the attribute name to map the datatype to, I dont get the right results back.
It seems like it should be `media.type`? But when I configure that as the attribute mapping, then go to the ui and select the datatype I want to search for.. it just disappears from the basic search page even though the advanced page is updated correctly and get expected result when running the search.

https://github.com/codice/ddf-ui/assets/14789712/59ea6ff2-b488-4d56-af53-74c1249f002a

kcwire:
Just pushed a new commit that should address the issue with the type disappearing.

pvargas (approved):

PR #846, opened 2023-10-31
Fix click-enter-to-save-geo race condition

alexabird:
The cesium and openlayers maps manage their location/drawing models separately. If both maps are open, hitting "enter" would cause both maps to save their locations which would be a race condition and sometimes the map without changes would overwrite the changes drawn on the other map. Its obvious that the changes should only be applied if the  there are changes on to the drawn geo, and that is what this PR fixes. Now, if both maps are open and only one map has been edited, the changes will be applied for the edited map.

Issues we will need to address:
- If the geo was edited on both maps, the race condition will still be an issue.
- If no changes happen, the UI doesnt do anything because it cant know what changes to apply.

Another issue this fixes:
Now that downstream projects are utilizes the drawing-and-display for openlayere (specifically the set color), the convertToModel only needs to specify the "color" attribute if it is not already set


andrewkfiedler (approved):

pvargas (approved):

jlcsmith:
build now

PR #847, opened 2023-10-31
Fix an issue where old map models were not properly cleared

andrewkfiedler:
- Added the more general search:drawend event as something to listen for when clearing drawing models
 - Updated the callback to compare against models, not just drawingModels, as the models are what exist due to the input being displayed somewhere in the dom (whereas filterModels are grabbed from the search).  The drawingModels are for when the user is actively drawing.


pvargas (approved):

sgalla (approved):

jlcsmith:
build now

PR #852, opened 2023-11-09
Drag and drop shapes on the 3D map

jrnorth:
https://github.com/codice/ddf-ui/assets/4495447/ee8bdc84-251a-47f8-93c4-5c4d2d5468ef

https://github.com/codice/ddf-ui/assets/14789712/febd62dc-8030-4774-9e79-1b0a896c5c06

Additional enhancements:
- The snack API now supports an `id` property, which can be used to replace a snack message with the same ID in the queue. Useful when several snack-generating events happen in quick succession, but you only want to display one.

Bug fixes:
- When the location input is cleared or removed, the corresponding shape should always be removed from the map now.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/interactions.provider.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/interactions.provider.tsx
@@ -1,0 +1,50 @@
+/* Copyright (c) Connexta, LLC */
+import React, { useState } from 'react'
+// @ts-expect-error ts-migrate(7016) FIXME: Could not find a declaration file for module 'cesi... Remove this comment to see the full error message
+import Cesium from 'cesium/Build/Cesium/Cesium'
+
+export type InteractionsContextType = {
+  interactiveGeo: number | null
+  setInteractiveGeo: (interactiveGeo: number | null) => void
+  moveFrom: Cesium.Cartesian3
+  setMoveFrom: (moveFrom: Cesium.Cartesian3 | null) => void
+  moveTo: Cesium.Cartesian3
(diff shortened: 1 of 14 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/interactions.provider.tsx
```diff
@@ -0,0 +1,60 @@
...
+  longitude: number
+  latitude: number
+}
+
+export type InteractionsContextType = {
+  interactiveGeo: number | null
+  setInteractiveGeo: (interactiveGeo: number | null) => void
+  interactiveModels: Backbone.Model[]
+  setInteractiveModels: (models: Backbone.Model[]) => void
+  moveFrom: Cesium.Cartographic | null
```
jrnorth: I've hardcoded Cesium in here for now, but we should be able to make this context generic so we can reuse it for the 2D map.

pvargas (approved):

alexabird (approved):

alexabird:
build now

alexabird:
"double click" to go into edit mode not working.
hitting "enter" no longer applies drawn geo changes.
After drag n drop, the draw model is not updated with the new location:
https://github.com/codice/ddf-ui/assets/14789712/16ee55a4-39a6-463b-830e-3f314cce7dd4

jrnorth:
@alexabird Nice finds! They have been fixed.

jrnorth:
build now

alexabird:
Hero looks good 🎉 The only thing I found was that the keyword geos can go into interactive mode after initial display but once they are re-rendered they get the "not allowed" cursor on hover; think this might be a preexisting issue from double click if we want to address it separately

PR #853, opened 2023-11-14
Automatically open 2d map when attempting to drawing a location filter

kcwire:
If the 2D map is not already open, then it will automatically open it for the user when they attempt to draw a location filter.

https://github.com/codice/ddf-ui/assets/4631300/fae4fc16-29ee-48ea-ab8e-0b48c5aad2f2

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/cross-window-communication.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/cross-window-communication.tsx
@@ -11,6 +11,7 @@ import wreqr from '../../js/wreqr'
 import GoldenLayout from 'golden-layout'
 import { getRootColumnContent, rootIsNotAColumn } from './stack-toolbar'
 import { unMaximize } from '../../react-component/visualization-selector/visualization-selector'
+import { Visualizations } from '../../component/visualization/visualizations'
 
 /**
  *  The popin function in golden layout has issues, particularly when there is a single stack in the main window at root.
@@ -335,7 +336,9 @@ function useConsumeSubwindowLayoutChange({
     return () => {}
   }, [goldenLayout, isInitialized])
 }
-
+function findOpenlayers(item: any) {
+  return item.componentName === 'openlayers'
+}
 /**
  *  Notice that we are only forwarding events that start with 'search' for now, as these are drawing events.
  */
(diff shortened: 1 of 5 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/cross-window-communication.tsx
```diff
@@ -352,6 +355,44 @@ const useProvideWreqrEvents = ({
     (event: string, args: any, { doNotPropagate = false } = {}) => {
       if (goldenLayout && isInitialized) {
         if (event.startsWith('search') && !doNotPropagate) {
+          // Launch the 2D Map (openlayers) if it's not already open
```
andrewkfiedler: Thinking we should move this into another file, with it's own hook name, like `useVerifyMapExistsWhenDrawing`.  This file is meant for all the cross window communication stuff for popping out golden layout visuals and keeping them in sync.

andrewkfiedler (changes requested):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/golden-layout/cross-window-communication.tsx
alexabird: Should we avoid opening the map if the 3D map is already open?
Should we check if the 2D map is open in another window?
If the 2D map is minimized, would it be easy to maximize it?

kcwire: - Added a check for cesium as well as openlayers so it will now only launch openlayers if neither map is open.
- If the map is minimized the drawing tools are correctly added to that map, but there isn't an easy way to "maximize" that ContentItem, as currently when you want to maximize a visual you drag it to a position, we would be guessing what position and potentially disrupt the user's layout.
- Also will now find any map visuals open in a separate window.

kcwire: Refactored into a separated file / hook as suggested.

alexabird (approved):

alexabird:
build now

malmgrens4:
🔥🔥 Hero Successful! 🔥🔥


https://github.com/codice/ddf-ui/assets/17283825/d0eea4cf-f428-40f4-b23d-ca13809210c4

malmgrens4 (approved):

PR #854, opened 2023-11-14
Add configuration for required export attributes

alexabird:
Adds a new configuration for required export attributes. These attributes will always be displayed in the inspector. When managing the inspector attributes, the user will be able to change the order of attributes, but not be able to select the attribute to be moved to "hidden"

https://github.com/codice/ddf-ui/assets/14789712/de53bcd1-d974-4ef8-a362-383e758a9e9e

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
@@ -152,6 +152,9 @@ public class ConfigurationApplication implements SparkApplication {
 
   private Integer exportResultLimit = 1000;
 
+  private List<String> requiredExportAttributes =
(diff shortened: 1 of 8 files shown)


ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/config/ConfigurationApplication.java
```diff
@@ -152,6 +152,9 @@ public class ConfigurationApplication implements SparkApplication {
 
   private Integer exportResultLimit = 1000;
 
+  private List<String> requiredExportAttributes =
+      ImmutableList.of("title", "security.classification", "security.releasability");
```
kcwire: I may be misremembering but is it possible to create a metacard type that doesn't include the "security" attributes? If that's the case we probably don't want to include them in this default list.

kcwire (approved):

pvargas (approved):

alexabird:
build now

PR #858, opened 2023-12-11
Preserve Geo Buffer

pvargas:
Prevent default geo values from overwriting user-set buffer width and unit.

Testing:
Test all shapes that have a buffer property
1. Draw a new shape, or update an existing one
2. Before clicking apply in the draw toolbar, change the buffer width or unit values
3. Click apply in the draw toolbar
4. Verify the buffer set in step 2 persists

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/drawing-and-display.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/drawing-and-display.tsx
@@ -367,6 +367,32 @@ export const OpenlayersDrawings = ({
     drawingLocation = null
   }
 
+  const preserveAttributes = (drawingModel: any, drawingLocation: any) => {
+    // preserve buffer
+    const updatedDrawingLocation = drawingLocation
+    if (drawingShape === 'Line') {
+      const lineWidth = drawingModel.get('lineWidth')
+      const lineUnits = drawingModel.get('lineUnits')
+      if (lineWidth) {
+        updatedDrawingLocation.properties.buffer = lineWidth
+      }
+      if (lineUnits) {
+        updatedDrawingLocation.properties.bufferUnit = lineUnits
+      }
+    }
+    if (drawingShape === 'Polygon') {
+      const polygonWidth = drawingModel.get('polygonBufferWidth')
+      const polygonUnits = drawingModel.get('polygonBufferUnits')
+      if (polygonWidth) {
+        updatedDrawingLocation.properties.buffer = polygonWidth
+      }
+      if (polygonUnits) {
+        updatedDrawingLocation.properties.bufferUnit = polygonUnits
+      }
+    }
+    return updatedDrawingLocation
+  }
+
   // called when the user clicks apply during geo drawing
   const finishDrawing = () => {
     if (drawingLocation === null) {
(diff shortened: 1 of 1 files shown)


jrnorth:
If you're adding a brand-new shape and enter a buffer value before drawing, will the buffer be displayed while you are drawing before you click apply? I'm pretty sure it won't, and if doesn't, then it would be nice to fix that as part of this PR.

alexabird:
build now

alexabird (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/drawing-and-display.tsx
```diff
@@ -50,6 +50,7 @@ import {
 } from '../../../../react-component/location/location-color-selector'
 import { InteractionsContext } from '../interactions.provider'
 import { Model } from 'backbone'
+import { Position } from '@blueprintjs/core'
```
jrnorth: Appears to be unused.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/drawing-and-display.tsx
```diff
@@ -170,12 +171,17 @@ const createGeoModel = (geo: GeometryJSON) => {
...
 
 const modelToPolygon = (model: any): GeometryJSON | null => {
-  const coords = model.get('polygon')
+  let coords = model.get('polygon')
   if (
     coords === undefined ||
     validateGeo('polygon', JSON.stringify(coords))?.error
   ) {
-    return null
+    coords = [
```
jrnorth: ❓Does this cause something to be displayed on the map at 0,0 when you start drawing?

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/drawing-and-display.tsx
```diff
@@ -379,7 +411,15 @@ export const OpenlayersDrawings = ({
     )
     wreqr.vent.trigger(`search:drawend`, drawingModel)
     drawingModel.set('drawing', false)
-    drawingModel.set(convertToModel(drawingLocation, drawingShape))
+
+    // preserve buffer set by user
+    const updatedDrawingLocation = preserveAttributes(
```
jrnorth: ✏️ Consider renaming the function to `preserveBuffer`, since that is more specific.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/drawing-and-display.tsx
```diff
@@ -379,7 +411,15 @@ export const OpenlayersDrawings = ({
...
+      drawingModel,
+      drawingLocation,
+      drawingShape
+    )
+
+    drawingModel.set(convertToModel(updatedDrawingLocation, drawingShape))
     setIsDrawing(false)
     setDrawingGeometry(drawingLocation)
```
jrnorth: ✏️ Recommend passing `updatedDrawingLocation` here as well. It doesn't make any functional difference, since it points to the same object, but it currently looks weird to use `updatedDrawingLocation` above, but not here.

jrnorth (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/openlayers/drawing-and-display.tsx
pvargas: You're right. The map would zoom in into existing shapes after opening, so I always missed that. Good catch!

pvargas:
build now

PR #860, opened 2024-01-03
Fix issues with shape drawing on initialize

andrewkfiedler:
- Updates the Query model to set the correct cql / filterTree on both itself and the LazyResult object right at construction time, preventing weirdness that could arise from the slight delay we have today
 - Removes a timeout effect that would inadvertantly use old data and possible cause shapes to disappear from the map.
 - Updates the drawing / display component to useCallback


pvargas (approved):

alexabird (approved):

alexabird:
Hero Successful 

https://github.com/codice/ddf-ui/assets/14789712/4dd98595-f1c8-4752-8780-3b9a686bb290

alexabird:
build now

PR #868, opened 2024-01-16
Update preference syncing across windows

andrewkfiedler:
- If golden layout happened to be open during uploads, weirdness would occur as uploads would be replaced during preferences syncs.
 - This updates our consumption of preference events to be particular about ignoring events from the window emitting them.  Might be worth doing the same for other events we have in this file, if only for performance purposes.


alexabird (approved):

alexabird:
build now

pvargas (approved):

PR #874, opened 2024-02-05
fix unreadable search results when attribute is updated

zack-leidos:
**What does this PR do, and why?**
The purpose of this PR is to fix a bug. When a user updates a result attribute and clicks on the unfocused (hidden) results tab, the items in the tab become unreadable. The PR addresses the issue.

**How should this be tested?**
Use the drop a Jar method to test the PR on an MIO

In the frontend application, upload two files in the Create section in the left menu bar
Make sure Results and Inspector windows are together as tabs in one window
Go to the Search function in the left side menu bar
Search for Fast data source
Click on one of the files you uploaded, then go to Inspector
Modify releasability and click save
Go back to the Results tab
You should see the results are unreadable then they appear readable (before, the results would remain unreadable)

**PR Definition of Done**
See that results are readable when the result attribute is changed.

PR #875, opened 2024-02-05
Update measure function to listen for active tab changes

andrewkfiedler:
- Update the virtual result lists to respond to active tab changes and to ignore events when the measure would be 0.
 - This fixes an issue where when the tab with a virtual list is currently hidden experiences an attribute update, the next time the user shows that tab it would have the wrong measurements (0) and be unreadable.


andrewkfiedler:
these changes have been pulled into https://github.com/codice/ddf-ui/pull/874

PR #897, opened 2024-03-15
Delete deleted metacards as current user subject instead of admin

sgalla:
In the code for reverting deleted metacards, we were running the delete operation on the deleted metacard as system admin, which isn't necessary.

Testing:
Verify you can delete and restore a metacard.


sgalla:
build now

pvargas (approved):

malmgrens4 (approved):

rececoffin (approved):

jlcsmith (approved):

PR #899, opened 2024-03-22
Fix Dragging Over Antimeridian

pvargas:
This PR addresses a 2D map issue where if a shape were to be dragged over the anti-meridian, the coordinates would become out of bounds.

Testing:
- to enter drag&drop mode, left click once over a shape. This will cause the shape to change color, this means to have entered drag&drop mode, and you can click and hold over the  shape, and drag the mouse to move the shape. Single click on the map to exit drag&drop mode.
- test all possible shapes
- drag shapes across the anti-meridian on both sides of the map
- make sure the coordinates are not out of bounds
- you should be able to enter and exit drag&drop mode at any time after dragging a shape of the anti-meridian
- test the 3D map, to make sure nothing has been broken

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/map.view.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/map.view.tsx
@@ -338,25 +338,38 @@ const getLocation = (model: Backbone.Model, translation?: Translation) => {
         'mapWest'
       )
       if (translation) {
-        bbox.mapNorth += translation.latitude
-        bbox.mapSouth += translation.latitude
-        bbox.mapEast += translation.longitude
-        bbox.mapWest += translation.longitude
+        const [east, north] = translateCoordinates(
+          [bbox.mapEast, bbox.mapNorth],
+          translation
+        )
+        const [west, south] = translateCoordinates(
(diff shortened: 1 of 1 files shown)


rececoffin:
build now

sgalla (approved):

rececoffin (approved):

sgalla:
The hero looked good, except for one little issue. When I position a geo across the anti-meridian, I can’t grab or drag the shape from the right side of the anti-meridian (but I can from the portion of the shape that’s on the left side of the anti-meridian).

I'm not sure how much time we want to spend trying to address this issue on this PR or if we could merge as-is and address this issue in the future.

https://github.com/codice/ddf-ui/assets/8962302/911d73d9-5c76-4888-804a-1d355cf00e72

pvargas:
> The hero looked good, except for one little issue. When I position a geo across the anti-meridian, I can’t grab or drag the shape from the right side of the anti-meridian (but I can from the portion of the shape that’s on the left side of the anti-meridian).


Good catch. I'll look into it.

pvargas:
> I'm not sure how much time we want to spend trying to address this issue on this PR or if we could merge as-is and address this issue in the future.

We've decided to marge this PR as-is, and address any outstanding issues in the future.

PR #909, opened 2024-04-01
Update layer component save calls to use the custom savePreferences rather than the backbone default

andrewkfiedler:
- Our custom savePreferences passes certain options like withoutSet that were not being used since this component used the backbone default save instead.  This led to issues with flickering if the user changed layer settings too quickly.
 - Updated the ordering callback to call savePreferences, as previously this wasn't being done.  As a result, if the user only reordered layers and didn't adjust the alpha or show, the reorder wouldn't be saved.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/layers/layers.view.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/layers/layers.view.tsx
@@ -69,7 +69,7 @@ const LayersViewReact = () => {
   const containerElementRef = React.useRef<HTMLDivElement>(null)
   const saveCallback = React.useMemo(() => {
     return debounce(() => {
-      user.get('user>preferences').save()
+      user.get('user>preferences').savePreferences()
     }, 100)
   }, [])
   useListenTo(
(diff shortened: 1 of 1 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/layers/layers.view.tsx
```diff
@@ -97,7 +97,7 @@ const LayersViewReact = () => {
               }
             )
             user.get('user>preferences').get('mapLayers').sort()
-            // user.get('user>preferences').save()
+            user.get('user>preferences').savePreferences()
```
sgalla, line 100: Is this necessary since `saveCallback` is getting called when `mapLayers` changes? Also, it looks like this line was commented out before. If so, is the `savePreferences` in the reset to defaults onClick is also unnecessary?

sgalla: lol oops, never mind. I should have actually read the PR description 🤦‍♀️  My bad!

sgalla (approved):

alexabird (approved):

jlcsmith:
build now

andrewkfiedler:
build now

alexabird:
Hero successful 

https://github.com/codice/ddf-ui/assets/14789712/c3fd8faf-7561-4d07-935c-ac0a93efee84

PR #916, opened 2024-04-19
Adds a new component to make type searching more ergonomic for users

andrewkfiedler:
- We now use a special type component that is based on the extra json config being set.  See the datatypes.tsx file for details on the json structure.
 - The structure allows grouping values, assigning icons to groups, ordering of groups (through their order in the json), custom values for each group that map to custom attributes / values for those attributes, and a custom icon for each value within a group.
 - If the extra json has no group specified, it will fall back to the previous datatype / type selection behavior, albiet with the new custom component.
 - The custom component is also utilized in advance search.  We do this by keeping the details of the component within the filter tree structure itself, and only mapping to proper cql when we have to.
 - This allows users to search on multiple type values, while also using advanced boolean logic like negation, without having to wrap their head around what those types expand to.  Previously a basic type search could look overwhelming, and unusable when changed into the advanced form.
 - The new property is inserted as an actual metacard attribute type on the backend when sending the config to the UI, called reserved.basic-datatype.  The goal is to avoid clashing with real attribute types.  We set it to be hidden, so that it doesn't show up anywhere unless we want it to.
 - By doing this, we can also take advantage of the alias config, which we do to set a custom label on this selector in the UI.  By default it has the old label of Types (see the alias set in ComposeApplication).
 - The cql parser has been updated to handle expansion of this new filter type before the actual writing.
 - Reserved properties file is here to ensure no circular dependencies between files that need to know the names of these reserved properties.
 - Various tests have been added to the cql parser to make sure it can handle the new structure, even when values are present that haven't been configured.  For now, it drops those values (but includes any valid ones).
 - Tests have also been added to exercise that the new type dropdown is able to properly transform the configuration values into forms it can understand.
 - Tests have been added to the query basic view to try and prevent regressions.  For the most part these are converting to the component state and back to filter to ensure there isn't a loss of information.  The actual component state doesn't get tested, only that it can handle going back to the filter as expected, so we should be able to flexibly update the component and catch any regressions.
 - The new property that shows the dropdown in the advanced form gets slotted under "Special Attributes" which will appear under "Common" for now.

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/compose/ComposeApplication.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/compose/ComposeApplication.java
@@ -174,6 +174,13 @@ public class ComposeApplication implements SparkApplication {
     anyDate.put("multivalued", false);
     anyDate.put("hidden", true);
 
+    Map<String, Object> reservedBasicDatatype = new HashMap<>();
+    reservedBasicDatatype.put("id", "reserved.basic-datatype");
+    reservedBasicDatatype.put("type", "STRING");
+    reservedBasicDatatype.put("multivalued", false);
+    reservedBasicDatatype.put("hidden", true);
+    reservedBasicDatatype.put("alias", "Types");
+
     Map<String, Object> metacardType = new HashMap<>();
     metacardType.put("id", "metacard-type");
     metacardType.put("type", "STRING");
@@ -199,6 +206,7 @@ public class ComposeApplication implements SparkApplication {
     attributeMap.put("anyText", anyText);
     attributeMap.put("anyGeo", anyGeo);
     attributeMap.put("anyDate", anyDate);
+    attributeMap.put("reserved.basic-datatype", reservedBasicDatatype);
     attributeMap.put("metacard-type", metacardType);
     attributeMap.put("source-id", sourceId);
     attributeMap.put("cached", cached);
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/filter-builder/filter.structure.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/filter-builder/filter.structure.tsx
@@ -15,8 +15,11 @@
 import moment from 'moment-timezone'
 import { ValuesType } from 'utility-types'
 import CQLUtils from '../../js/CQLUtils'
+import { Omit } from '../../typescript'
+
 import { SpreadOperatorProtectedClass } from '../../typescript/classes'
 import ExtensionPoints from '../../extension-points'
+import { BasicDataTypePropertyName } from './reserved.properties'
 
 export const deserialize = {
   /**
@@ -164,6 +167,11 @@ export class FilterBuilderClass extends BaseFilterBuilderClass {
   }
 }
 
+/**
+ *  We want to support more complex negation than allowed by the normal filter tree, so we store negated in a property.
+ *  However, since the write function in cql.tsx doesn't know about this, at some point we need to convert it back to this class where negation exists as a type instead of property.
+ *  See the uncollapseNots function in cql.tsx for where this is done.
+ */
 export class CQLStandardFilterBuilderClass extends BaseFilterBuilderClass {
   declare readonly type: 'AND' | 'OR' | 'NOT'
   declare readonly filters: Array<
@@ -224,6 +232,7 @@ export type ValueTypes = {
     start: number
     end: number
   }
+  multivalue: string[]
   booleanText: BooleanTextType
   location: // this is all we technically need to reconstruct (lo fidelity)
   LineLocation | PolygonLocation | PointRadiusLocation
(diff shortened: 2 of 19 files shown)


ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/compose/ComposeApplication.java
```diff
@@ -174,6 +174,13 @@ private Map<String, Object> getMandatoryAttributes() {
     anyDate.put("multivalued", false);
     anyDate.put("hidden", true);
 
+    Map<String, Object> reservedBasicDatatype = new HashMap<>();
+    reservedBasicDatatype.put("id", "reserved.basic-datatype");
```
andrewkfiedler, line 178: This could be changed to whatever we want.  The goal would be to ensure we don't clash with real attribute names.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/compose/ComposeApplication.java
```diff
@@ -174,6 +174,13 @@ private Map<String, Object> getMandatoryAttributes() {
...
+    reservedBasicDatatype.put("type", "STRING");
+    reservedBasicDatatype.put("multivalued", false);
+    reservedBasicDatatype.put("hidden", true);
+    reservedBasicDatatype.put("alias", "Types");
```
andrewkfiedler, line 182: Default alias which acts as the new component's label.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/compose/ComposeApplication.java
```diff
@@ -174,6 +174,13 @@ private Map<String, Object> getMandatoryAttributes() {
     anyDate.put("multivalued", false);
     anyDate.put("hidden", true);
 
+    Map<String, Object> reservedBasicDatatype = new HashMap<>();
+    reservedBasicDatatype.put("id", "reserved.basic-datatype");
+    reservedBasicDatatype.put("type", "STRING");
+    reservedBasicDatatype.put("multivalued", false);
+    reservedBasicDatatype.put("hidden", true);
```
andrewkfiedler, line 181: Ensures this doesn't show up anywhere in the UI unless we want it to.

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/compose/ComposeApplication.java
```diff
@@ -174,6 +174,13 @@ private Map<String, Object> getMandatoryAttributes() {
     anyDate.put("multivalued", false);
     anyDate.put("hidden", true);
 
+    Map<String, Object> reservedBasicDatatype = new HashMap<>();
+    reservedBasicDatatype.put("id", "reserved.basic-datatype");
+    reservedBasicDatatype.put("type", "STRING");
```
andrewkfiedler, line 179: Type and multivalued end up not really mattering, since we key off the id to do our custom components.  So essentially these are just filler values.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/query-basic/query-basic.view.spec.tsx
```diff
@@ -0,0 +1,548 @@
...
+    })
+  }
+  return obj
+}
+
+function removeCustomClasses(obj: any) {
+  return JSON.parse(JSON.stringify(obj))
+}
+
+describe('verify going back and forth from filter to state in query basic is not lossy or "wrong"', () => {
```
andrewkfiedler, line 44: I wanted to make sure we didn't test the internals of this component so that we could refactor it later freely and still have the tests pass.

Essentially this tests is we can give it a value, and then ask for the value back without it losing any information.
Our state => component state => our state

ui-frontend/packages/catalog-ui-search/src/main/webapp/js/IconHelper.tsx
```diff
@@ -149,15 +149,6 @@ export default {
     const i = _deriveIconByMetacardObject(metacard)
     return _get(i, 'class', _default.class)
   },
-  getUnicode() {
```
andrewkfiedler, line 152: Unused

andrewkfiedler:
build now

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/filter-builder/filter.structure.tsx
```diff
@@ -164,6 +167,11 @@ export class FilterBuilderClass extends BaseFilterBuilderClass {
   }
 }
 
+/**
+ *  We want to support more complex negation than allowed by the normal filter tree, so we store negated in a property.
+ *  However, since the write function in cql.tsx doesn't know about this, at some point we need to convert it back to this class where negation exists as a type instead of property.
+ *  See the uncollapseNots function in cql.tsx for where this is done.
+ */
 export class CQLStandardFilterBuilderClass extends BaseFilterBuilderClass {
```
malmgrens4, line 175: For my understanding, could you break down what's going on here a little more. 
What does negation as a type vs property mean?

andrewkfiedler: Yeah, I left this comment to refresh myself as well.  Basically a while back, maybe a few years at this point, folks wanted to do negation in a less complex way in the UI than cql actually allowed.  Essentially doing negation on leafs, not just entire branches.  That meant making our own version of what a filter tree could be.  

So this is a class to differentiate when we move to the cql official filter tree, that the parser can actually understand and translate to cql.  I'm not sure if this helps.  Let me know.

andrewkfiedler: Screenshot 2024-04-19 at 6 41 02 PM


Pic might help.  It's so we can do it on fields, not just groups.  But since that isn't officially supported, we at some point transform to the officially supported format (which surrounds it in a group).

andrewkfiedler:
build now

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/filter-builder/filter.structure.tsx
malmgrens4: yep very helpful, ty

malmgrens4: so at some point we turn what the UI has as a field into a group (which can be negated). 

So for clarity translating to something like this isn't possible/ undesired behavior `-field_txt:"value"`

malmgrens4:
https://github.com/codice/ddf-ui/assets/17283825/547d63cb-cbf9-4b6d-9d26-d5dde344a47e

Looks like the facets need to be reset/ converting from other types of matches

malmgrens4:
Search works as expected. Negating the fields works as expected.
Mixed with title search for testing more complex queries.

malmgrens4:
When switching filter between types of matches CONTAINS, NEAR, etc. and to/from the entity/event subtypes it looks like the dropdown's value needs to be reset.

malmgrens4 (approved):

malmgrens4:
Tested filters with different versions of the provided config. Searches worked based on the fields in the configs

alexabird (approved):

pvargas (approved):

PR #918, opened 2024-05-14
Updates various drawing utilities to enable better downstream capabilities, fix bugs

andrewkfiedler:
- Adds a unique id to locations models to ensure trackability.  In the future we should update other code areas to defer to id instead of locationId, if possible.
 - Updates resizable grid to allow controlling it from a parent, for better dev ux
 - Updates the interactions provider to mark the models that are interactive as interactive, so that it's simpler to find out what is being interacted with.  In the future, we could remove the interactions provider and instead have the existing code that is using it mark the models itself.
 - Removes the unused map.ts component, and updates how we export the openlayers / cesium interfaces so their types are accessible, for better dev ux
 - Updates usage of zoom to home so that it's solely triggered within the drawing / display component, so now we appropriately will go to either the home location or the shapes drawn on the map.  Before, we would end up always going to the home.
 - Updates cesium map component so it can pass the map upwards, similar to the openlayers map.  This is useful for parent components that want to do some action once the map loads.
 - Adds a constant to the cesium map interactions component for simpler panning / zooming of the camera to a proper orientation (LookingStraightDown).
 - Fixes panToShapesExtent so that it will properly zoom / pan to the shapes
 - Adds zoomToIds to both map interfaces for zooming / panning to particular shapes.
 - Updates drawing / display to slightly update convertToModel, allowing passing in an existing model to keep certain properties around.
 - Updates drawing / display for openlayers props to match the one for cesium again, as there is no longer a need to pass in drawing / display models from a parent.
 - Similar update to openlayers component to eliminate the drawing / display models prop.
 - Remove of zoom to home from components that no longer need it.


andrewkfiedler:
build now

andrewkfiedler:
build now

alexabird (approved):

pvargas (approved):

rececoffin (approved):

alexabird:
(tested in a downstream project so links wont be available to those without access)

The only things that seem isolated to these changes:
* After “Applying” a drawn shape, the draw color is retained instead of the location's color

  https://github.com/connexta/aus/assets/14789712/c4404fd5-1269-440b-ab0b-5aadcfb5f1a1

* The extra verbiage about might needing to save the item seem unnecessary
verbiage


Some other things I noticed that could be improvements:
* Go into draw mode and hit ENTER - nothing happens. If you edit the shape and hit ENTER, it applies the changes. If you go into draw mode and hit ESC, it cancels. Seems like hitting enter when there are no changes should effectively cancel.
* On drag&drop, hitting ESCAPE will exit interactive mode, hitting ENTER does nothing. 
* The drag&drop hover icons can get into a bad state. While in “interactive” mode, User must hover over the shape to start drag, but the “grab” cursor shows on non-hover and disappears if you happen to not click on the geo when initiating the drag. While dragging, the cursor will go back-n-forth between ”pointer” and “grabbing”.

  https://github.com/connexta/aus/assets/14789712/ba638260-117f-4181-9cec-5403d22eea9a

andrewkfiedler:
> (tested in a downstream project so links wont be available to those without access)
> 
> The only things that seem isolated to these changes:
> 
>     * After “Applying” a drawn shape, the draw color is retained instead of the location's color
>       
>         
>           
>           
>       
>           Screen.Recording.2024-05-15.at.9.07.01.PM.mov
> 
>     * The extra verbiage about might needing to save the item seem unnecessary
>       verbiage
> 
> 
> Some other things I noticed that could be improvements:
> 
>     * Go into draw mode and hit ENTER - nothing happens. If you edit the shape and hit ENTER, it applies the changes. If you go into draw mode and hit ESC, it cancels. Seems like hitting enter when there are no changes should effectively cancel.
> 
>     * On drag&drop, hitting ESCAPE will exit interactive mode, hitting ENTER does nothing.
> 
>     * The drag&drop hover icons can get into a bad state. While in “interactive” mode, User must hover over the shape to start drag, but the “grab” cursor shows on non-hover and disappears if you happen to not click on the geo when initiating the drag. While dragging, the cursor will go back-n-forth between ”pointer” and “grabbing”.
>       
>         
>           
>           
>       
>           Screen.Recording.2024-05-15.at.8.31.38.PM.mov

Pushed up a fix for that draw color issue, had the logic on the if statement reversed.

The other improvements would be nice to get in at some point.  I think I can see how we could update the hover / cursor logic to fix it, but I'd rather separate that into a different pr.  

A side note but I noticed we have some bugs in main (unrelated to the pr) around editing shapes on maps when they're popped out.  Shapes on non popped out maps get out of sync with those popped out (or vice versa).

PR #920, opened 2024-05-20
Move determination of hidden to be in central area within compose, fix sorting regression

andrewkfiedler:
- The previous move of sorting to the backend did not take into account casing.  Now we properly do a case insensitive sort.
 - The logic around hidden attributes was previously split between configuration / metacard definitions on the frontend.  This logic has been moved to be done upfront, similar to the sorting / alias mapping, on the backend instead.  Now we can check for the presence of hidden on the attribute definitions.
 - With this in mind, added hidden to the anyText, anyGeo definitions on the backend.  These are now manually included in the one attribute list we want them in rather than risking them leaking elsewhere like the metacard editor.  See getFilteredAttributeList in filterHelper.


rececoffin (approved):

jlcsmith:
build now

PR #921, opened 2024-05-20
Adds the ability to override download handling downstream

andrewkfiedler:
- Downstream projects can now provide their own components to handle downloads.
 - The snack function now returns a function to remove the snack if the component calling it desires.
 - Small update to the useOverridable hook to allow function based overrides (in this case, a download component aka function).


andrewkfiedler:
build now

jlcsmith:
build now

PR #928, opened 2024-06-12
Keep Exisiting Shapes In 3D Map While Drawing a New Shape

pvargas:
In order to align the behaviors of the 2D and 3D maps, while in draw mode, the old shapes in a 3D map will now remain on the map until the user applies or cancels the changes.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
@@ -61,6 +61,39 @@ export const removeOldDrawing = ({ map, id }: { map: any; id: string }) => {
   relevantPrimitives.length > 0 && map.getMap().scene.requestRender()
 }
 
+const makeOldShapeNonEditable = ({ map, id }: { map: any; id: string }) => {
+  const relevantPrimitives = map
+    .getMap()
+    .scene.primitives._primitives.filter((prim: any) => {
+      return prim.id === id
+    })
+  relevantPrimitives.forEach((relevantPrimitive: any) => {
+    if (typeof relevantPrimitive.setEditMode === 'function') {
+      relevantPrimitive.setEditMode(false)
+    }
+    if (typeof relevantPrimitive.setEditable === 'function') {
+      relevantPrimitive.setEditable(false)
+    }
+  })
+  relevantPrimitives.length > 0 && map.getMap().scene.requestRender()
+}
+
+export const removeOrLockOldDrawing = (
+  isInteractive: boolean,
+  id: any,
+  map: any,
+  model: any
+) => {
+  if (
+    isInteractive ||
+    (!isInteractive && Object.keys(model.changed).includes('isInteractive'))
+  ) {
+    removeOldDrawing({ map, id })
+  } else {
+    makeOldShapeNonEditable({ map, id })
+  }
+}
+
 let drawingLocation: any
 
 const updateDrawingLocation = (newDrawingLocation: any) => {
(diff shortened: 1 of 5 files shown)


alexabird:
build now

pvargas:
build now

pvargas:
build now

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
```diff
@@ -61,6 +61,123 @@ export const removeOldDrawing = ({ map, id }: { map: any; id: string }) => {
...
+    if (typeof relevantPrimitive.setEditable === 'function') {
+      relevantPrimitive.setEditable(false)
+    }
+  })
+  relevantPrimitives.length > 0 && map.getMap().scene.requestRender()
+}
+
+const nestedArraysOverlap = (arrayA: any[], arrayB: any[]) => {
+  let result = false
+  arrayA.map((elemA) => {
```
sgalla: Does this need to be a map function? The map will iterate over every element in arrayA, but it seems like it just cares that any element has overlap with any element of arrayB. I might not be understanding it correctly, but it seems like this would work?

```suggestion
  return arrayA.some((elemA) => arrayB.some((elemB) => JSON.stringify(elemA) === JSON.stringify(elemB)))
```

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
```diff
@@ -61,6 +61,123 @@ export const removeOldDrawing = ({ map, id }: { map: any; id: string }) => {
...
+          west: prevModel['west'],
+          south: prevModel['south'],
+        }
+        if (prevBox.north) {
+          return !(
+            box.north === prevBox.north ||
+            box.east === prevBox.east ||
+            box.west === prevBox.west ||
+            box.south === prevBox.south
+          )
```
sgalla, line 111: Should these be ANDed together instead of ORed? Maybe I'm not understanding what exactly this logic should be doing, but seems like logic for determining if the bbox is different should be any coordinate not equal, so `north !== north || south !== south ||... `, and the equivalent logic for would be `!(north === north && south === south &&...)`  

Edit: Okay I see similar logic for circle, so it's considered a new shape when ALL coordinates are different, not just one.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
pvargas: I think that would work. Thanks!

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
pvargas: Yep, your edit is correct. The negation is just so the value matches the name of the function.

sgalla (approved):

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/visualization/maps/cesium/drawing-and-display.tsx
```diff
@@ -61,6 +61,123 @@ export const removeOldDrawing = ({ map, id }: { map: any; id: string }) => {
...
+
+const nestedArraysOverlap = (arrayA: any[], arrayB: any[]) => {
+  let result = false
+  return arrayA.some((elemA) => arrayB.some((elemB) => JSON.stringify(elemA) === JSON.stringify(elemB)))
+    result =
+      result ||
+      arrayB.some((elemB) => JSON.stringify(elemA) === JSON.stringify(elemB))
+  })
+  return result
+}
```
sgalla: Oops, I didn't do that suggestion right I guess... I think this would work though:
```suggestion
const nestedArraysOverlap = (arrayA: any[], arrayB: any[]) => {
  return arrayA.some((elemA) => arrayB.some((elemB) => JSON.stringify(elemA) === JSON.stringify(elemB)))
}
```

alexabird (approved):

sgalla:
build now

sgalla:
build now

PR #933, opened 2024-06-17
Fallback to user preferences for 2D map settings

sgalla:
Fall back to user preferences to store map layers and coordinate format settings for the 2D map when it is not in the context of a layout. This allows the openlayers map component to be used outside the context of a golden layout.


sgalla:
build now

andrewkfiedler (approved):

alexabird (approved):

alexabird:
Used `"catalog-ui-search": "https://github.com/codice/ddf-ui#catalog-ui-search-dist-933",` in a downstream project and verified that the openlayers map that was used outside of the GoldenLayout renders correctly 🎉

jlcsmith:
build now

PR #934, opened 2024-06-17
Refactor search index tracking to enable advanced paging / resume scenarios

andrewkfiedler:
- There is a desire to be able to resume searches at any point, so rather than track the previous indexes the logic around paging was updated to be able to resume using the current index alone.  This required some rework to depend only on current index + status.
 - To ensure no regressions, tests have been refactored to extensively cover each edge case for paging, including a mix of federated + local sources.  During this process, several bugs with the previous logic were fixed, mostly around these edges.
 - Updated the query methods to enable the new use cases around resuming a search.  There is now a canRefetch and refetch method on query.
 - The paging logic in query methods is also refactored, with the correct results verified through extensive testing.  Some of this logic could likely be refactored to be more straightforward, in particular the next page vs constrained next page, but for now the tests at least verify it's working, and will allow us to refactor at a later date with confidence.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/search.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/search.tsx
@@ -6,7 +6,6 @@ import {
 } from '../resizable-grid/resizable-grid'
 import SelectionInterfaceModel from '../selection-interface/selection-interface.model'
 import { useQuery, useUserQuery } from '../../js/model/TypedQuery'
-import Grid from '@mui/material/Grid'
 import Paper from '@mui/material/Paper'
 import { QueryAddReact } from '../query-add/query-add'
 import KeyboardArrowLeftIcon from '@mui/icons-material/KeyboardArrowLeft'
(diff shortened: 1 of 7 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/pages/search.tsx
```diff
@@ -1439,6 +1428,8 @@ export const HomePage = () => {
     queryModel: selectionInterface.getCurrentQuery(),
     on: searchPageMode === 'adhoc',
   })
+  // console.log('queryModel', selectionInterface.getCurrentQuery())
+  // console.log('queryModelJSON', selectionInterface.getCurrentQuery().toJSON())
```
sgalla: Can these be removed?

sgalla (approved):

andrewkfiedler:
build now

malmgrens4 (approved):

malmgrens4:
Tested with remote and local sources. 
Sources reached their "max" and locked appropriately. 
Paging throughout the app behaved as expected. 
Page size is respected. Changes to hit counts is respected.
Queries can be interrupted. 

Screenshot 2024-06-18 at 2 03 54 PM
Screenshot 2024-06-18 at 2 03 47 PM

PR #939, opened 2024-06-21
Fix source maps

andrewkfiedler:
- Updates our dist to include the original source so downstream projects using webpack can correctly process source maps and debug if necessary.

Explanation of publish command
1.  Remove all contents of the dist directory to ensure a fresh build environment 
2.  Copy non-TypeScript source files from src to dist/src
3. Copy all .ts and .tsx files from the project, excluding node_modules, dev, and target directories, to dist/_src-map-dir/
4. Copy all files from the current directory to dist
5. Compile the TypeScript files in the dist/_src-map-dir with an output dir of /dist
6. Copy the package.json file to dist and set its private field to false


andrewkfiedler:
build now

PR #944, opened 2024-06-25
Update to use inline source maps

andrewkfiedler:
- Theoretically we could ship separate source but that would be more effort for minimal payout.  Downstream projects and even the app itself will not become larger due to inline source maps, as terser strips them out in production (they are comments) when bundling.  I've verified this downstream, just to be sure.


andrewkfiedler:
build now

pvargas (approved):

PR #946, opened 2024-06-27
Add null check for when hitCountPerSource value is null

alexabird:


jlcsmith (approved):

ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/util/CqlQueriesImpl.java
```diff
@@ -152,7 +152,9 @@ public CqlQueryResponse executeCqlQuery(CqlRequest cqlRequest)
           (key, value) -> {
             if (key.startsWith(METRICS_SOURCE_ELAPSED_PREFIX)) {
               String source = key.substring(METRICS_SOURCE_ELAPSED_PREFIX.length());
-              elapsedPerSource.put(source, new Long((Integer) value));
+              if (value != null) {
```
glenhein: A slightly better check would be `value instanceof Integer`. If value is null, then this will return false. It also protects against a possible cast error.

glenhein (approved):

alexabird:
build now

andrewkfiedler (approved):

PR #947, opened 2024-06-27
Escape Attribute Name Symbols in Boolean Query

pvargas:
This PR formats the the property/attribute name in a boolean search by escaping non-alphanumeric characters and delimiting the string with quotes. This change prevents boolean queries containing non-alphanumeric characters from failing.

To test, perform an advanced search with
- an attribute containing at least one symbol (`-` or `.`)
- the boolean operator
- a valid search term

The query should return results as expected.

Example search 
Screenshot 2024-06-27 at 15 52 45

### DIFF
--- a/ui-backend/boolean-search/boolean-search-application/src/main/java/org/codice/ddf/catalog/search/handlers/QueryHandler.java
+++ b/ui-backend/boolean-search/boolean-search-application/src/main/java/org/codice/ddf/catalog/search/handlers/QueryHandler.java
@@ -28,6 +28,11 @@ public class QueryHandler implements Handler {
     // implementation not needed
   }
 
+  public static String formatSearchProperty(String str) {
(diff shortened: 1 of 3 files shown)


pvargas:
build now

alexabird (approved):

jlcsmith (approved):

andrewkfiedler (approved):

andrewkfiedler:
I think this is a good fix, but we might want to consider updating the endpoint to handle the escaping.  

If we update https://github.com/codice/ddf-ui/blob/master/ui-backend/boolean-search/boolean-search-application/src/main/java/org/codice/ddf/catalog/search/handlers/QueryHandler.java#L34 I think we wouldn't have to worry about this on the frontend, as it looks the query param is being sent correctly.

pvargas:
build now

andrewkfiedler (approved):

PR #949, opened 2024-07-01
set default source querySetting value from all to local

LiamYokoyama:
This change is to ensure that the default value for search sources that is selected is local(fast sources).
When 'all' is selected as default, search is perceived to be "slow", due to the searching including slow sources by default.

To test.
1. Build catalog-ui jar and drop into /ddf/deploy
2. Verify that your user has no preference metacards in solr. e.g delete the metacard that is associated with query: metacard-tags_txt:ddf-preferences && user_txt:<your_user>
3. On the landing page, verify that the default source drop down is defaulting to "fast sources" and that the metacard that gets created (using step 2 query) has 'local' in 'querySettings'

### DIFF
--- a/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/preferences/DefaultCatalogUiPreferences.java
+++ b/ui-backend/catalog-ui-search/src/main/java/org/codice/ddf/catalog/ui/preferences/DefaultCatalogUiPreferences.java
@@ -95,7 +95,7 @@ public class DefaultCatalogUiPreferences implements DefaultPreferencesSupplier {
         "querySettings",
         new ImmutableMap.Builder<>()
             .put("type", "text")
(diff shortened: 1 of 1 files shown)


LiamYokoyama:
build now

LiamYokoyama:
build now

andrewkfiedler (approved):

malmgrens4 (approved):

jlcsmith:
@LiamYokoyama  - shouldn't this be handled by the "Default Sources" property in the Catalog UI Search config?

alexabird:
> shouldn't this be handled by the "Default Sources" property in the Catalog UI Search config?

It sounds like there's work underway to utilize the Catalog UI Search config values as the user's defaults here (there are a couple of values other than "sources" that also aren't using the Catalog UI Search config), and this is a "quick fix" for sources.

PR #954, opened 2024-07-11
Download Latest Resource Version

pvargas:
This change prevents cached resources from being downloaded. This is useful when users have overwritten a resource.

To test:
1. Upload a product
2. Overwrite that product with something else
3. Download the product
4. Repeat steps 2 and 3 a few times and verify the most recent version is always downloaded.


pvargas:
build now

sgalla (approved):

alexabird (approved):

PR #957, opened 2024-07-16
Update query set to ensure filterTree is always the correct type

andrewkfiedler:
This ensures that regardless of the set / initialize we'll always be able to know a query.get('filterTree') is of the correct type.

Previously we were only catching instances of string in the set, thinking the defaults area would handle this conversion, however that turned out to not always be the case.


sgalla (approved):

alexabird (approved):

alexabird:
build now

alexabird:
build now

alexabird:
Not sure why the build keeps aborting - I was able to get a successful `mvn clean install` locally

PR #958, opened 2024-07-17
Move cookie to be set as soon as possible on app start

andrewkfiedler:
- Certain connections only get the value of the cookie at their opening, like websockets.  To ensure we don't have issues around a missing cookie, we can move it here and ensure that the moment we know we need a cookie we apply it.


malmgrens4 (approved):

sgalla (approved):

glenhein (approved):

glenhein:
Hero successful.

andrewkfiedler:
Build now

PR #962, opened 2024-07-24
Split up cql query by source for current page export

malmgrens4:
We were sending all the ids on the current page as the query. This would break when you had a large number of results from several sources (e.g. 2 remote sources with 250 results and your local source you end up with a query of 750 ids being federated) 
This packages the requests by source so the most ids sent to any one source will be the max page size (250). The endpoint was made to handle this case we just weren't using it.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/table-export/table-export.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/table-export/table-export.tsx
@@ -42,6 +42,7 @@ import ProgressButton from '../../react-component/progress-button/progress-butto
 import DialogContent from '@mui/material/DialogContent/DialogContent'
 import DialogActions from '@mui/material/DialogActions/DialogActions'
 import DialogContentText from '@mui/material/DialogContentText'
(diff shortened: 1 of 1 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/table-export/table-export.tsx
```diff
@@ -151,25 +154,47 @@ export const getExportBody = async (ExportInfo: ExportInfo) => {
...
+    const resultIdSourcePairs: SourceIdPair[] = Object.values(
+      query.get('result').get('lazyResults').results
+    ).map((result: LazyQueryResult) => ({
+      id: result.plain.metacard.properties['id'],
+      sourceId: result.plain.metacard.properties['source-id'],
+    }))
 
-  const searches = [
-    {
+    console.log(resultIdSourcePairs)
```
malmgrens4: delete

jlcsmith (approved):

pvargas (approved):

jlcsmith:
build now

derekwilhelm (approved):

PR #963, opened 2024-07-24
[5.1.26.x] Split cql by source on export

malmgrens4:
We were sending all the ids on the current page as the query. This would break when you had a large number of results from several sources (e.g. 2 remote sources with 250 results and your local source you end up with a query of 750 ids being federated) 
This packages the requests by source so the most ids sent to any one source will be the max page size (250). The endpoint was made to handle this case we just weren't using it.


derekwilhelm (approved):

pvargas (approved):

alexabird:
build now

jlcsmith (approved):

PR #965, opened 2024-07-26
Allow unsuccessful results to be counted

derekwilhelm:
Some federated sources can produce error details but still return results.  Currently, when the UI receives results from a query for a source that indicates an error, the available and possible counts reported do not include updates from "unsuccessful" sources.  This means that if the source did in fact return results, they would be displayed to the user in the results visual but the status in the heartbeat area would indicate that there were zero results.  This is a confusing state for the user to be in.

This PR removes the filtering of unsuccessful results so that all of the returned data is counted appropriately.


jlcsmith (approved):

glenhein (approved):

sgalla (approved):

andrewkfiedler (approved):

derekwilhelm:
build now

pvargas (approved):

malmgrens4 (approved):

derekwilhelm:
build now

PR #966, opened 2024-07-29
Update filter components to support non metacard based filters, add better number validation

andrewkfiedler:
- It's useful to use these filter components on non metacard filters in order to build UI components that hit solr for other things, which this will allow.
 - The number validation on some of the filters was preventing users from doing things like deleting their value and starting with a negative, so it's been rewritten to keep values around locally no matter what, and only propagate those as changes to the filter when they're valid.  If unvalid, the user is notified that the previous valid value will be used instead.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-around.spec.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-around.spec.tsx
@@ -113,10 +113,11 @@ describe('verify date around field works', () => {
     verifyDateRender('12', 'minute', data.date1.userFormat12.minute)
   )
   it('calls onChange with updated value when precision changes', () => {
+    let updatedDate = data.date1.userFormatISO.millisecond
     wrapper = mount(
       <DateAroundField
         value={{
-          date: data.date1.userFormatISO.millisecond,
+          date: updatedDate,
           buffer: {
             amount: '1',
             unit: 'd',
@@ -124,7 +125,7 @@ describe('verify date around field works', () => {
           direction: 'both',
         }}
         onChange={(updatedValue) => {
-          expect(updatedValue.date).to.equal(data.date1.utcISOMinutes)
+          updatedDate = updatedValue.date
         }}
       />
     )
@@ -132,5 +133,8 @@ describe('verify date around field works', () => {
       .get('user')
       .get('preferences')
       .set('dateTimeFormat', Common.getDateTimeFormats()['ISO']['minute'])
+    setTimeout(() => {
+      expect(updatedDate).to.equal(data.date1.utcISOMinutes)
+    }, 100)
   })
 })
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-around.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-around.tsx
@@ -59,6 +59,10 @@ const validateDate = (
 }
 
 export const DateAroundField = ({ value, onChange }: DateAroundProps) => {
+  const validValue = {
+    ...defaultValue(),
+    ...value,
+  }
   const dateRef = React.useRef(value.date)
   const blueprintDateRef = React.useRef<DateInput>(null)
 
@@ -70,8 +74,7 @@ export const DateAroundField = ({ value, onChange }: DateAroundProps) => {
       DateHelpers.Blueprint.converters.UntimeshiftFromDatePicker(shiftedDate)
     dateRef.current = unshiftedDate.toISOString()
     onChange({
-      ...defaultValue(),
-      ...value,
+      ...validValue,
       date: unshiftedDate.toISOString(),
     })
   })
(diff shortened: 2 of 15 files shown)


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-around.spec.tsx
```diff
@@ -113,24 +113,28 @@ describe('verify date around field works', () => {
...
-          expect(updatedValue.date).to.equal(data.date1.utcISOMinutes)
+          updatedDate = updatedValue.date
         }}
       />
     )
     user
       .get('user')
       .get('preferences')
       .set('dateTimeFormat', Common.getDateTimeFormats()['ISO']['minute'])
+    setTimeout(() => {
```
andrewkfiedler, line 136: Avoids the issue where we generate an onChange before the expected onChange around precision.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-around.tsx
```diff
@@ -59,6 +59,10 @@ const validateDate = (
 }
 
 export const DateAroundField = ({ value, onChange }: DateAroundProps) => {
+  const validValue = {
```
andrewkfiedler, line 62: Move the handling of optional / invalid values up to this common point rather than down in the component.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-around.tsx
```diff
@@ -137,20 +133,16 @@ export const DateAroundField = ({ value, onChange }: DateAroundProps) => {
...
+                    amount: val.toString(),
                   },
                 })
             }}
-            {...(value.buffer
-              ? {
-                  value: value.buffer.amount,
-                }
-              : {})}
+            validation={(val) => val > 0}
```
andrewkfiedler, line 143: Notice we can pass in extra validation to the number field, in this case to say we only allow positive numbers.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/date-relative.tsx
```diff
@@ -26,6 +26,10 @@ const isInvalid = ({ value }: Props) => {
 }
 
 export const DateRelativeField = ({ value, onChange }: Props) => {
+  const validValue = {
```
andrewkfiedler, line 29: Similar change to move handling of invalid values to the top of the component.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/number.tsx
```diff
@@ -2,35 +2,106 @@ import * as React from 'react'
...
+}: Props) {
+  const [localValue, setLocalValue] = React.useState<LocalValueType>(
+    parseValue(value || '1', type)
+  )
+  const [hasValidationIssues, setHasValidationIssues] = React.useState(false)
+  const [constructedValidationText, setConstructedValidationText] =
+    React.useState('')
+
+  React.useEffect(() => {
+    if (
```
andrewkfiedler, line 47: Here is where we decide if we will send the new value to the filter (or other parent component with an onChange).

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/filter-builder/filter.structure.util.tsx
```diff
@@ -0,0 +1,221 @@
+import { FilterClass, ValueType, ValueTypes } from './filter.structure'
+
+export const isRelativeValue = (
```
andrewkfiedler, line 3: A bunch of utilities  to make it possible to narrow down what type of filter you're handling.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/filter-builder/filter.structure.util.tsx
```diff
@@ -0,0 +1,221 @@
...
+    typeof value === 'object' &&
+    value !== null &&
+    'type' in value &&
+    (value.type === 'LINE' ||
+      value.type === 'POLYGON' ||
+      value.type === 'POINTRADIUS')
+  )
+}
+
+export const isValueEmpty = (filter: FilterClass): boolean => {
```
andrewkfiedler, line 135: A general utility around deciding if a filter has been "filled out" so to speak and is valid.

ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/filter/filter-comparator/comparatorUtils.tsx
```diff
@@ -149,3 +152,38 @@ export const getComparators = (attribute: string): ComparatorType[] => {
   }
   return comparators
 }
+
+export const ComparatorContext = React.createContext({
```
andrewkfiedler, line 156: This is what enables us to properly assign comparators in non metacard situations.

andrewkfiedler:
build now

PR #967, opened 2024-07-30
Update drawer components to include the forward / backward nav if desired

andrewkfiedler:
- These allow users to more seamlessly use the included navigation buttons when the drawer is involved in their history.


ui-frontend/packages/catalog-ui-search/src/main/webapp/react-component/user-settings/user-settings.tsx
```diff
@@ -35,7 +35,7 @@ const ThemeGlobalStyle = createGlobalStyle`
   opacity: 0 !important;
 }
 .MuiDrawer-root > .MuiPaper-root {
-  transform: scale(.8) translateY(40%) translateX(-10%) !important;
```
andrewkfiedler, line 38: Fixes a longstanding issue where the theme UI would be slightly offscreen on certain window sizes

alexabird (approved):

pvargas (approved):

andrewkfiedler:
build now

PR #975, opened 2024-08-19
Support exporting deleted items with Export As interaction

sgalla:


sgalla:
build now

sgalla:
build now

jlcsmith (approved):

pvargas (approved):

andrewkfiedler:
Is there a downside to adding a metacard.tags * to those cql filters?  It might be more simple.  Something like:
```
 {
              type: 'ILIKE',
              property: '"metacard-tags"',
              value: '*',
}
```

andrewkfiedler:
> Is there a downside to adding a metacard.tags * to those cql filters? It might be more simple. Something like:
> 
> ```
>  {
>               type: 'ILIKE',
>               property: '"metacard-tags"',
>               value: '*',
> }
> ```

We do something like this to ensure we get a result back when refreshing result data:
https://github.com/codice/ddf-ui/blob/master/ui-frontend/packages/catalog-ui-search/src/main/webapp/js/model/QueryResult.ts#L200

sgalla:
build now

zkirksey:
Hero was successful!

PR #976, opened 2024-08-20
Export overwrite url as parameter

malmgrens4:
Allows the overwrite component to have a custom URL. 
Since the component isn't being used with a custom URL here this should be tested by making sure there is no regression with the ddf-ui overwrite function.


jlcsmith (approved):

jlcsmith:
build now

andrewkfiedler (approved):

malmgrens4:
any opposition to merging this?

PR #979, opened 2024-08-27
Add better size filtering, expand filter class to add context property

andrewkfiedler:
- Size filtering is now possible if the configuration for resourceSizeIdentifiers is set within extraJson configuration for the UI. In addition, the context provider can be used to provide this set, in case of folks wanting a custom filter experience.
 - The size filtering is done through a new class where value is unchanged, but context as a new property is available.  This allows us to more easily add this change in, as we don't have to go through and change value types / handle that sort of migration.
 - If context is available, it will be used, otherwise we fallback appropriately, in both directions.
 - The more human understandable interpretation of sizes is used, with 1000 as the pow instead of 1024.
 - Refactors the cql methods for handling more complex filters to be a bit more efficient by only going through the tree a single time.

 - When the resource size components render, they check shape.  If the shape is incorrect, they will attempt to reset with a debounce of 250ms.  In the meantime, they will show a loading indicator.


ui-frontend/packages/catalog-ui-search/src/main/webapp/component/fields/resource-size-range.tsx
```diff
@@ -0,0 +1,228 @@
...
+  value: unknown,
+  defaultValue: number
+): number {
+  try {
+    const castedValue = value as number
+    if (!isNaN(castedValue)) {
+      return castedValue
+    }
+  } catch (e) {
+    // If casting fails, return null to indicate such (default depends on start or end)
```
sgalla: This comment isn't accurate, is it?

andrewkfiedler: I forget to fix that after changing this to instead take in a default value, thanks!

sgalla (approved):

andrewkfiedler:
build now

andrewkfiedler:
build now

jlcsmith (approved):

PR #981, opened 2024-08-29
Get default sources from sources from configuration instead of using "local"

glenhein:
To test, clear Solr of def-preferences metacards, or use a user that doesn't have a preference metacard. In the admin catalog-ui system settings, set the default sources. Then log in as the user. In Settings->Search, make sure the sources are pre-selected based on the configuration. Go to Saved Searches and click the New Search button. Make sure the search has the correct sources pre-selected. Pull up the ddf-preferences metacard and make sure querySettings_txt has the sources as expected.


andrewkfiedler (approved):

pvargas (approved):

glenhein:
build now

jlcsmith:
@glenhein what happens with users that have existing ddf-preferences?  Will their default sources be based on their preferences?

PR #982, opened 2024-09-09
Fixes an issue where if a filter starts with an incorrect value type (such as string), it will always appear to be loading

andrewkfiedler:


andrewkfiedler:
build now

PR #984, opened 2024-09-12
Update how elevated rights works, making it purely cookie based

andrewkfiedler:
- There were issues around this being in preferences since it involves solr.
 - This moves the elevated rights mode to be done purely through the cookie, as the UX is nearly identical but more stable


andrewkfiedler:
build now

PR #988, opened 2024-09-24
Update fonts to use fallbacks immediately, then swap

andrewkfiedler:
- This might help in case certain environments load custom fonts slower than expected.
 - Fixes an issue in theme where we create a theme without a font specified and merge it into one with it specified.  In this case, it was causing the font default to be correct, but individual components to use the default Mui fonts.


derekwilhelm (approved):

jlcsmith (approved):

jlcsmith:
build now

alexabird (approved):

andrewkfiedler:
build now

PR #995, opened 2024-11-13
Updated array check for cross window functionality

malmgrens4:
Cross window array checks would fail - for example in the inspector pop out - because the instanceof check is dependent on the window.Array constructor being the same as the current window's. See more details below:

From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray:

Array.isArray() checks if the passed value is an Array. It does not check the value's prototype chain, nor does it rely on the Array constructor it is attached to. It returns true for any value that was created using the array literal syntax or the Array constructor. This makes it safe to use with cross-realm objects, where the identity of the Array constructor is different and would therefore cause instanceof Array to fail.

### DIFF
--- a/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/app/app.tsx
+++ b/ui-frontend/packages/catalog-ui-search/src/main/webapp/component/app/app.tsx
@@ -96,7 +96,7 @@ const matchesRoute = ({
     )
   } else if (
     routeInfo.routeProps.path &&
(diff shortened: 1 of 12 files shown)


andrewkfiedler:
Nice!! 🚀

andrewkfiedler (approved):

alexabird (approved):

alexabird:
build now

PR #998, opened 2024-12-12
Fix the display of multivalue  attributes in the results

alexabird:
The results panel would crash if GEO or FLOAT attributes had multiple values. 

To test:
1. Create a product with >1 extracted location
2. "Manage attributes" so that the extracted locations attribute displays
3. Confirm the correct information is displayed in the results panel


alexabird:
build now

zkirksey (approved):

jlcsmith (approved):

PR #1011, opened 2025-02-19
Update MUI theme so dialogs have no max width

andrewkfiedler:
- We previously did this in a global styles css file which no longer exists.  This is a better place for it, as it uses the MUI theme object instead.


andrewkfiedler:
build now

jlcsmith (approved):

alexabird (approved):

PR #1014, opened 2025-02-25
Update dialogs to instead use maxWidth lg, rather than setting it to none

andrewkfiedler:
- There were some changes in mui v6 that are hard to pin down, but amongst them is dialog classes / styles changed a bit.  Previously it seemed like setting no max width would be a good solution, but it messes with too many dialogs.  A better middle ground appears to be setting the maxWidth for the dialog to lg for now.


andrewkfiedler:
build now

alexabird (approved):

malmgrens4 (approved):

PR #1016, opened 2025-03-02
Fix issues with multiple react router instances for react router v6

andrewkfiedler:
- In v6, react router moved to an internal history object, which means navigation in separate roots no longer send events to the original router.
 - We now render a single router at the root of the app rather than in providers.
 - We have a hook that runs within that router to sync context to an outside variable (history and location).
 - The non primary react roots now are wrapped in a provider that uses that to construct history and location contexts.
 - The provider also handles golden layout subwindows, as those need to handle navigation events by emitting them back to the main window.


andrewkfiedler:
build now

alexabird (approved):

andrewkfiedler:
build now

malmgrens4 (approved):

PR #1029, opened 2025-03-17
Migrate editable attributes / can write permission extensions to overridable pattern

andrewkfiedler:
- Migrates these extension points to use overridable instead as it's more flexible, and involves less need to entangle them within startup procedures.
 - Updates the can write usages to use isWritable rather than isNotWritable as that's easier for folks to understand.
 - Fixes an issue with the original editable attributes extension in the new overridable, as previously this could be called many times resulting in a massive amount of network calls, when in reality it needs to only be called once.


alexabird (approved):

jlcsmith:
build now

jlcsmith (approved):

PR #1030, opened 2025-03-19
Update menu state to auto wrap popover usages with the menu state context provider

andrewkfiedler:
- This will automatically handle the majority of use cases, as most times in the application we are directly feeding popover props to a popover.  Now those components within will automatically get wrapped with the relevant menu state context.
 - For other uses, we provide MenuStateProviderInstance as a convenience for wrapping components with context.
 - Updates some of the function calls to utilize useCallback when possible, to minimize rerenders.
 - Adds some default implementations for the context functions that will show warnings to devs in the console if they happen to be used in a situation where menu state context is missing.


andrewkfiedler:
build now

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/menu-state/menu-state.tsx
```diff
@@ -84,20 +86,73 @@ const useListenForChildUpdates = ({
...
   const anchorRef = React.useRef<HTMLDivElement>(null)
   const popoverRef = useRerenderingRef<HTMLDivElement>()
   const action = React.useRef<PopoverActions | null>(null)
   const [open, setOpen] = React.useState(false)
   useListenForChildUpdates({ popoverRef: popoverRef.current, action })
-  const handleClick = () => {
-    setOpen(!open)
-  }
+  const handleClick = React.useCallback(() => {
+    setOpen((currentValue) => {
```
andrewkfiedler, line 125: Updated this one to a callback and used the ability to access previous value so we could keep the dependency array blank

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/menu-state/menu-state.tsx
```diff
@@ -109,13 +164,19 @@ export function useMenuState({ maxHeight }: Props = {}) {
...
     },
+    // these can also be used directly with the MuiPopper if desired
     MuiPopoverProps: {
       open,
       onClose: handleClose,
       anchorEl: anchorRef.current,
       action,
       ref: popoverRef.ref,
       ...POPOVER_DEFAULTS({ maxHeight }),
+      slotProps: {
```
andrewkfiedler, line 186: In most cases we use these popover props, splatting them into the popover, so we'll now automatically get the menu context in those cases.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/menu-state/menu-state.tsx
```diff
@@ -137,7 +199,10 @@ export function useMenuState({ maxHeight }: Props = {}) {
       ref: anchorRef,
       onClick: handleClick,
     },
+    MenuStateProviderInstance,
```
andrewkfiedler, line 213: For other use cases not using popovers, we have this as a convenience for wrapping things with the menu context.

ui-frontend/packages/catalog-ui-search/src/main/webapp/component/app/react-router.patches.tsx
```diff
@@ -28,7 +28,6 @@ class SubscribableNavigationContextClass extends Subscribable<{
   ) {
     this.navigationContext = navigationContext
     this._notifySubscribers({ thing: 'update' })
-    console.log('updateNavigationContext', navigationContext)
```
andrewkfiedler, line 31: Removing some chatty console logs that we don't need.

alexabird (approved):

malmgrens4 (approved):

andrewkfiedler:
build now

PR #1037, opened 2025-03-21
Fix issues with openlayer geometry styles sometimes not appearing correctly

andrewkfiedler:
- This aligns the addLine method to be more like other areas involving styles that use the proper get / set methods that exist on Features in openlayers.  Previous versions it doesn't appear to have mattered, but with the latest upgrade it appears to cause issues if we don't use them, likely due to how they manage the Feature object internally.

Current issue (seen with results that have complex geometries):
Screenshot 2025-03-21 at 3 25 02 PM
Screenshot 2025-03-21 at 3 25 07 PM

With the fix:
Screenshot 2025-03-21 at 3 25 41 PM
Screenshot 2025-03-21 at 3 25 37 PM


alexabird (approved):

glenhein (approved):

alexabird:
build now

PR #1040, opened 2025-03-31
Fix the logic for updating old drawings vs locking them

andrewkfiedler:
- In the old logic, translating a shape on another map (click, then drag) would cause other cesium maps to show both the new location of the shape and the old.
 -  Now, each shape directly calls the appropriate method (lock or remove) based on internal branching logic, preventing duplicate renderings and ensuring consistent state across maps.

Here is an example of the old behavior, where upon moving the shape on the top map (openlayers), the bottom map cesium shows two shapes.
Screenshot 2025-03-31 at 2 41 48 PM


andrewkfiedler:
build now

andrewkfiedler:
build now

alexabird (approved):

jlcsmith:
build now

jlcsmith (approved):

PR #1041, opened 2025-04-01
[5.2.9.x] Fix the logic for updating old drawings vs locking them

andrewkfiedler:
- In the old logic, translating a shape on another map (click, then drag) would cause other cesium maps to show both the new location of the shape and the old.
 -  Now, each shape directly calls the appropriate method (lock or remove) based on internal branching logic, preventing duplicate renderings and ensuring consistent state across maps.
 
 Backport of https://github.com/codice/ddf-ui/pull/1040


andrewkfiedler:
build now

andrewkfiedler:
build now

PR #1043, opened 2025-04-01
[5.2.9.x] Updates the date range overlap test to properly diff the dates

andrewkfiedler:
- The previous version would fail if run on the final day of the month, since the next day would be smaller than it (resulting in something like -30 on the 31st instead of 1).  This uses moment so the diff is always going to work.

backport of https://github.com/codice/ddf-ui/pull/1042


andrewkfiedler:
build now

PR #1049, opened 2025-04-07
[5.2.9.x] Update timeline to setHeight when initializing

andrewkfiedler:
- Sometimes the initial resize is missed due to race conditions, so we need to make sure we call this independently from the resize listener at least once.


alexabird (approved):

andrewkfiedler:
build now

andrewkfiedler:
build now

jlcsmith (approved):

PR #1050, opened 2025-04-07
Update timeline to setHeight when initializing

andrewkfiedler:
- Sometimes the initial resize is missed due to race conditions, so we need to make sure we call this independently from the resize listener at least once.


jlcsmith (approved):

jlcsmith:
build now

alexabird (approved):

PR #1069, opened 2025-08-04
Flatten getGeometries to handled multivalue geo attributes

alexabird:
Previously clustering would fail to include results that had multivalued geo attributes
image

Now, all geometry attributes are used to calculate a center point for the result metacard. This center point is used to determine if the result belongs to a cluster.

Future improvement consideration: Instead of calculating a center point per result metacrd, treat each geometry on the metacard as its own geo when determining clusters


alexabird:
build now

andrewkfiedler (approved):

derekwilhelm (approved):

PR #1072, opened 2025-08-23
Update openlayers layer handling to properly handle wmts, and always wrapX

andrewkfiedler:
- Updates openlayers layer handling to verify layer identity / matrix set identity to avoid errors being thrown.  If these are invalid, it will use the first of each.
 - Updates wmts layers for openlayers to always use wrapX, as the default openlayers behavior is too strict for this to be determined automatically.


andrewkfiedler:
build now

alexabird (approved):

malmgrens4 (approved):

PR #1073, opened 2025-08-25
Pan to the extent of all geometries on a result

alexabird:
Instead of only panning to the location attribute, get all geometry attributes on the metacard


andrewkfiedler (approved):

alexabird:
build now

malmgrens4 (approved):

PR #1088, opened 2025-12-12
Use defaultValue for customInput

malmgrens4:
Value is not state in this component. When value is updated by a parent rerenders can cause the cursor to jump to end of input field.


alexabird (approved):

alexabird:
build now

jlcsmith (approved):

PR #1151, opened 2026-06-08
Fix idle-timeout logout

smithmlw:
The session-timeout components in catalog-ui-search and admin were navigating
directly to /session/invalidate (browser redirect), so the Keycloak logout URL
returned in the response body was never used. The browser reloaded the current
page, the OIDC profile was still present, and the session was never properly
torn down through Keycloak, leaving users on "Logout Failed!" after idle expiry.

Change both components to fetch /session/invalidate and navigate to the URL
returned as response text, matching the flow the manual logout button already
used. Encode the return-to URL before appending it. Extract the Spark handler
into a named method for testability and add a test.

### DIFF
--- a/ui-backend/catalog-ui-search/src/test/java/org/codice/ddf/catalog/ui/session/SessionManagementApplicationTest.java
+++ b/ui-backend/catalog-ui-search/src/test/java/org/codice/ddf/catalog/ui/session/SessionManagementApplicationTest.java
@@ -13,41 +13,36 @@
  */
 package org.codice.ddf.catalog.ui.session;
 
-import static spark.Spark.get;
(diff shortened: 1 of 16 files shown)


derekwilhelm (approved):

.gitignore
```diff
@@ -17,4 +17,6 @@ dist/
 yarn-debug.log*
 yarn-error.log*
 ddf-ui-deplist.out
-**/.java-version
\ No newline at end of file
+**/.java-version
+tmp/
+build_20*.log
```
jlcsmith: 📝 probably can be removed

smithmlw: removed

alexabird (approved):

